PR #17, opened 2018-10-25
Access repos via https in INSTALL.md

killthekitten:
It is simpler and requires less effort (i.e. no auth) compared to SSH


soumith:
thanks @killthekitten !

Issue #31, opened 2018-10-26
RuntimeError: The shape of the mask [230202] at index 0 does not match the shape of the indexed tensor [0] at index 0


liminghao1630:
## ❓ Questions and Help
Traceback (most recent call last):
  File "github/maskrcnn-benchmark/tools/train_net.py", line 170, in <module>
    main()
  File "github/maskrcnn-benchmark/tools/train_net.py", line 163, in main
    model = train(cfg, args.local_rank, args.distributed)
  File "github/maskrcnn-benchmark/tools/train_net.py", line 73, in train
    arguments,
  File "/home/v-minghl/github/maskrcnn-benchmark/maskrcnn_benchmark/engine/trainer.py", line 65, in do_train
    loss_dict = model(images, targets)
  File "/home/v-minghl/anaconda3/lib/python3.7/site-packages/torch/nn/modules/module.py", line 477, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/v-minghl/github/maskrcnn-benchmark/maskrcnn_benchmark/modeling/detector/generalized_rcnn.py", line 50, in forward
    proposals, proposal_losses = self.rpn(images, features, targets)
  File "/home/v-minghl/anaconda3/lib/python3.7/site-packages/torch/nn/modules/module.py", line 477, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/v-minghl/github/maskrcnn-benchmark/maskrcnn_benchmark/modeling/rpn/rpn.py", line 94, in forward
    return self._forward_train(anchors, objectness, rpn_box_regression, targets)
  File "/home/v-minghl/github/maskrcnn-benchmark/maskrcnn_benchmark/modeling/rpn/rpn.py", line 113, in _forward_train
    anchors, objectness, rpn_box_regression, targets
  File "/home/v-minghl/github/maskrcnn-benchmark/maskrcnn_benchmark/modeling/rpn/loss.py", line 91, in __call__
    labels, regression_targets = self.prepare_targets(anchors, targets)
  File "/home/v-minghl/github/maskrcnn-benchmark/maskrcnn_benchmark/modeling/rpn/loss.py", line 62, in prepare_targets
    labels_per_image[~anchors_per_image.get_field("visibility")] = -1
RuntimeError: The shape of the mask [230202] at index 0 does not match the shape of the indexed tensor [0] at index 0

Hi,
[...]

PR #37, opened 2018-10-26
Raise error if no GT or proposals available during training

fmassa:
This should help catch problems similar to the one in https://github.com/facebookresearch/maskrcnn-benchmark/issues/31

### DIFF
--- a/maskrcnn_benchmark/modeling/matcher.py
+++ b/maskrcnn_benchmark/modeling/matcher.py
@@ -51,9 +51,15 @@ class Matcher(object):
             be matched.
         """
         if match_quality_matrix.numel() == 0:
-            # handle empty case
-            device = match_quality_matrix.device
-            return torch.empty((0,), dtype=torch.int64, device=device)
+            # empty targets or proposals not supported during training
+            if match_quality_matrix.shape[0] == 0:
+                raise ValueError(
+                    "No ground-truth boxes available for one of the images "
+                    "during training")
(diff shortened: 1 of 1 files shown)


maskrcnn_benchmark/modeling/matcher.py
```diff
@@ -51,9 +51,15 @@ def __call__(self, match_quality_matrix):
             be matched.
         """
         if match_quality_matrix.numel() == 0:
-            # handle empty case
-            device = match_quality_matrix.device
-            return torch.empty((0,), dtype=torch.int64, device=device)
+            # empty targets or proposals not supported during training
+            if match_quality_matrix.shape[0] == 0:
+                raise ValueError(
+                    "No ground-truth boxes available for one of the images "
```
soumith: make it tell the user which image / filename if possible. otherwise it becomes a bit of a nightmare to track back

fmassa:
Propagating which image yielded the empty target can be a bit tricky as there might have been random crops etc.

Maybe I'll add another assert like this in the Dataset itself, but in the meantime I'll be merging this as is.

maskrcnn_benchmark/modeling/matcher.py
fmassa: I agree as well, but for the image names to be displayed in the error message means that the assert would need to be applied in the user code as well, when they create a new Dataset class

PR #40, opened 2018-10-26
Use dist.get_rank() instead of local_rank to detect master process

bermanmaxim:
Current implementation is subject to [PyTorch issue 12042](https://github.com/pytorch/pytorch/issues/12042) due to using `--local-rank` to detect the master process when training on more than 1 nodes.

This fix uses `distributed.get_rank()` to detect the master process and should therefore allow training with multiple nodes on shared file systems.

### DIFF
--- a/maskrcnn_benchmark/utils/comm.py
+++ b/maskrcnn_benchmark/utils/comm.py
@@ -18,6 +18,12 @@ def get_world_size():
     return torch.distributed.deprecated.get_world_size()
 
 
+def get_rank():
+    if not torch.distributed.deprecated.is_initialized():
+        return 0
+    return torch.distributed.deprecated.get_rank()
+
+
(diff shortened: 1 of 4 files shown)


bermanmaxim:
You're welcome. Actually just realized I could have used [`maskrcnn_benchmark/utils/comm.is_main_process`](https://github.com/bermanmaxim/maskrcnn-benchmark/blob/2e4f03b122bce8f8dce61e3371ebaf5b5e292434/maskrcnn_benchmark/utils/comm.py#L27) as it was already defined as a wrapper around `get_rank() == 0`. Not sure if it's worth doing that stylistic change now that it's merged 🤔

fmassa:
There are a number of those changes that can be applied. If you batch a few of them, don't hesitate to send another PR!

PR #41, opened 2018-10-26
Fix BoxList constructor call in README.md

killthekitten:


soumith:
thanks @killthekitten !

PR #70, opened 2018-10-30
Make pretrained model dir customizable

martinruenz:
Training machines might not have an internet connection and it can be useful to customize the location of pretrained models. For example, training machines could share a network attached drive for these models. 

To this end, this change adds the setting `PRETRAINED_DIR`.

### DIFF
--- a/maskrcnn_benchmark/config/defaults.py
+++ b/maskrcnn_benchmark/config/defaults.py
@@ -267,3 +267,4 @@ _C.TEST.IMS_PER_BATCH = 8
 _C.OUTPUT_DIR = "."
 
 _C.PATHS_CATALOG = os.path.join(os.path.dirname(__file__), "paths_catalog.py")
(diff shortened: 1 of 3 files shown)


fmassa:
Hi,

Thanks for the PR!

So, let me see if I understand it properly: you want a way of making the pre-trained models directory customizable by the user, is that right?

For that, we have the possibility to pass a `PATHS_CATALOG` python file to the training / testing script, which allows the user to change all the paths to be something else.

Plus, we can also use the environment variable `TORCH_MODEL_ZOO` to change the location where the models will be saved / loaded.

Is there any other use-case that I'm missing that you'd want in `maskrcnn-benchmark`?

martinruenz:
Thanks for the clarification! Somehow I missed `TORCH_MODEL_ZOO` when looking at `cache_url`, which indeed makes this pull request redundant.

PR #78, opened 2018-10-31
adds Dockerfile

miguelvr:
Added Dockerfile to build with CUDA

Example build script with defaults (`CUDA_VERSION=9.0`, `CUDNN_VERSION=7`):
```bash
nvidia-docker build -t maskrcnn-benchmark docker/ 
```
Example build script with other CUDA and CUDNN versions:

```bash
nvidia-docker build -t --build-arg CUDA_VERSION=9.2 --build-arg CUDNN_VERSION=7 maskrcnn-benchmark docker/ 
```

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -10,7 +10,7 @@
 - (optional) OpenCV for the webcam demo
 
 
-### Step-by-step installation
+### Option 1: Step-by-step installation
 
 ```bash
 # first, make sure that your conda is setup properly with the right environment
@@ -55,3 +55,13 @@ python setup.py build develop
 # or if you are on macOS
 # MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py build develop
 ```
+
+### Option 2: Docker Image (Requires CUDA)
+
+Build image with defaults (`CUDA_VERSION=9.0`, `CUDNN_VERSION=7`):
+
+    nvidia-docker build -t maskrcnn-benchmark docker/
+    
+Build image with other CUDA and CUDNN versions:
+
+    nvidia-docker build -t --build-arg CUDA_VERSION=9.2 --build-arg CUDNN_VERSION=7 maskrcnn-benchmark docker/ 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -56,7 +56,7 @@ python setup.py build develop
 # MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py build develop
 ```
 
-### Option 2: Docker Image (Requires CUDA)
+### Option 2: Docker Image (Requires CUDA, Linux Only)
 
 Build image with defaults (`CUDA_VERSION=9.0`, `CUDNN_VERSION=7`):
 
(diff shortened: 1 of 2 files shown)


miguelvr:
Sure thing! You're welcome, thanks for the awesome project!

robpal1990:
> Sure thing! You're welcome, thanks for the awesome project!

This looks great. Could you add one which does not use conda/miniconda as well?

miguelvr:
> > Sure thing! You're welcome, thanks for the awesome project!
> 
> This looks great. Could you add one which does not use conda/miniconda as well?

why would you want that? It makes no difference because you're in a docker container. 

I just used conda because I know it is the preferred method to install pytorch due to the MKL version.

soumith:
yea i dont think non-conda based Dockerfile is needed.

soumith:
If there's a way for one dockerfile to start from another dockerfile, we should use this as the base Dockerfile https://github.com/pytorch/pytorch/blob/master/docker/pytorch/Dockerfile

miguelvr:
> If there's a way for one dockerfile to start from another dockerfile, we should use this as the base Dockerfile https://github.com/pytorch/pytorch/blob/master/docker/pytorch/Dockerfile

You can start a docker from an image, not another dockerfile. If you have an image built and hosted on a dockerhub I would be able to do it. But I am seeing that pytorch's image is for CUDA 9.0 only.

robpal1990:
> > If there's a way for one dockerfile to start from another dockerfile, we should use this as the base Dockerfile https://github.com/pytorch/pytorch/blob/master/docker/pytorch/Dockerfile
> 
> You can start a docker from an image, not another dockerfile. If you have an image built and hosted on a dockerhub I would be able to do it. But I am seeing that pytorch's image is for CUDA 9.0 only.

You can always paste the contents of the dockerfile you created at the end of the pytorch dockerfile that Soumith linked here. Not elegant but should work.

miguelvr:
@fmassa please check if those instructions are enough

fmassa:
I think this looks good, but I'll let @soumith comment as well because I don't have much experience with Docker.

mingruimingrui:
The current Image would be built without NCCL which is needed for fast distributed training.
Might I suggest running 
`apt-get update && apt-get install -y --no-install-recommends libnccl2=2.3.5-2+cuda9.0`
under `# install basics`>

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
+ARG CUDA_VERSION=9.0
+ARG CUDNN_VERSION=7
+
+FROM nvidia/cuda:${CUDA_VERSION}-cudnn${CUDNN_VERSION}-devel-ubuntu16.04
+
+# install basics
+RUN apt-get update -y \
+ && apt-get install -y git curl ca-certificates bzip2 cmake tree htop bmon iotop g++
```
fmassa: don't you need an extra `\` in the end of this line?

INSTALL.md
```diff
@@ -55,3 +55,13 @@ python setup.py build develop
 # or if you are on macOS
 # MACOSX_DEPLOYMENT_TARGET=10.9 CC=clang CXX=clang++ python setup.py build develop
 ```
+
+### Option 2: Docker Image (Requires CUDA)
```
fmassa: Could you also add a note saying that it doesn't work with OSX?

miguelvr: sorry, didn't see this. why doesn't it work in OSX btw?

PR #102, opened 2018-11-03
Add RetinaNet Implementation

chengyangfu:
Hi, 
    This PR contains the RetinaNet implementation. The following table contains the models which use ResNet50 and ResNet101 as the backbones. 

GPU : Pascal Titan X
PyTorch : v1.0.0 
Inference time is measured when setting batch size as 1. 

|  Model       |  Detectron Accuracy |  Current Accuracy  | inference time(s/im) | download |
| ------------- |:-------------:|:-----:|:-------:|:----:|
| RetinaNet_R-50-FPN_1x | 35.7 | 36.3 | 0.102 |  [model](http://www.cs.unc.edu/~cyfu/retinanet_R-50-FPN_1x_model.pth) |
| RetinaNet_R-101-FPN_1x | 37.7 | 38.5 | 0.123 | [model](http://www.cs.unc.edu/~cyfu/retinanet_R-101-FPN_1x_model.pth) |
| RetinaNet_X-101-32x8d-FPN_1x | 39.5 | 39.8 | 0.200 | [model](http://www.cs.unc.edu/~cyfu/retinanet_X-101-32x8d-FPN_1x_model.pth) |  
| RetinaNet_R-50-FPN_P5_1x | 35.7 | 36.2 | 0.097 |   [model](http://www.cs.unc.edu/~cyfu/retinanet_R-50-FPN_P5_1x_model.pth) |
| RetinaNet_R-101-FPN_P5_1x | 37.7 | 38.5 | 0.121 | [model](http://www.cs.unc.edu/~cyfu/retinanet_R-101-FPN_P5_1x_model.pth) |

Add  _C.TEST.DETECTIONS_PER_IMG = 100. 
After using DETECTIONS_PER_IMG, the mAP drops 0.1. 

Not Implemented parts.  
-  Class specific bbox prediction. 
-  Softmax Focal Loss

---------------------------------------------------------------------------------------------------
Updated 02/02/2018
Identify the reason why this branch gets higher AP.

| Branch | Accuracy | Difference | 
| -------- | ------------ | ----- |
| This | 36.3/55.2/38.9/</br>19.7/39.9/49.0  | BoxCoder(10, 10, 5, 5), </br> add *4 in classification loss normalization |
| [retinanet-detectron](https://github.com/chengyangfu/maskrcnn-benchmark/tree/retinanet_detectron) | 35.6/55.8/37.7/</br>19.6/39.3/48.2 | BoxCoder(1, 1, 1, 1) | 

---------------------------------------------------------------------------------------------------
Updated 01/30/2018
After updating PyTorch to  v1.0.0, the inference time reduced around 15~20%. 
Update the inference time in the table. 

--------------------------------------------------------------------------------------------------
Updated 01/26/2018
Add RetinaNet_X-101-32x8d-FPN_1x model.
AP : 39.8 
Inferece time : 0.200 second.

---------------------------------------------------------------------------------------------------
Updated 01/25/2018
In my first version, I accidentally used P5 to generate P6 instead of C5 which was used in Detectron and paper. 
The following table compares the performances in these two settings. 

| Model | C5            |   P5 |
| -------- | ------------ | ----- |
| RetinaNet_R-50-FPN_1x | 36.3/55.2/38.9/19.7/39.9/49.0 | 36.2/55.1/38.7/19.7/39.5/48.6 |
| RetinaNet_R-101-FPN_1x | 38.5/57.6/41.0/20.8/42.3/51.7 | 38.5/57.9/41.3/21.0/42.8/51.3 |

---------------------------------------------------------------------------------------------------
Updated 01/23/2018

Train the model without ["divide by 4"](https://github.com/chengyangfu/maskrcnn-benchmark/blob/retinanet/maskrcnn_benchmark/modeling/rpn/retinanet/loss.py#L145) in the regression loss. 
Performance: 

| Model            | AP | AP50 | AP75 | APs | APm | APl |
| ------------- | --- | ----- | ------ | --- | ------ | ---- |
| RetinaNet_R-50-FPN_1x | 29.6 | 45.0 | 31.5 | 13.9 | 31.4 | 41.2 | 

---------------------------------------------------------------------------------------------------
Updated 11/20/2018

The matching part is slightly different from the Detectron version. 
In Detectron [matching](  https://github.com/facebookresearch/Detectron/blob/master/detectron/roi_data/retinanet.py#L219-L231), anchors with IOU >= 0.5 are considered as positive examples, and anchors with IOU <=0.4 are negative examples. Then for the low qualities matches(best prediction for each gt), Detectron only uses the low-quality examples >= 0.4.  

P.S.In [Detectron](https://github.com/facebookresearch/Detectron/blob/master/detectron/roi_data/retinanet.py#L228-L229), there are some cases occur in both fg_inds and bg_inds. Although in Line230, Detectron removes all the low-qualities positive examples < 0.4. I think the in Line231, num_fg calculation is not correct. <br>
I also test the threshold used for low-qualities positive examples from 0.5 to 0.0.   

| threshold | AP | AP50 | AP75 | P.S. |
| --- | --- | --- | --- | --- |
| 0.5 | 35.5 | 53.7 | 38.1 | |
| 0.4 | 36.0 | 54.1 | 38.6 | Detectron version. |
| 0.3 | 36.1 | 54.5 | 38.9 | |
| 0.2 | 36.1 | 54.5 | 38.7 | | 
| 0.0 | 36.2 | 55.0 | 38.7 | This branch. |


fmassa:
This is really awesome, thanks a lot for the PR!

I'll have a closer look at it next week, let us know the result of the training!

chengyangfu:
Finishing the training. 
RetinaNet with X_101_32x8d backbone model costs too much time for training now. Due to the CVPR submission deadline is coming, our lab does not have extra machines for training this one. If anyone can train this one, I will highly appreciate it!

fmassa:
No worries about X_101_32x8d training, we can do it on our side.

maskrcnn_benchmark/config/defaults.py
```diff
@@ -221,6 +221,75 @@
 _C.MODEL.RESNETS.RES2_OUT_CHANNELS = 256
 _C.MODEL.RESNETS.STEM_OUT_CHANNELS = 64
 
+
+# ---------------------------------------------------------------------------- #
+# RetinaNet Options (Follow the Detectron version)
+# ---------------------------------------------------------------------------- #
+_C.RETINANET = CN()
```
fmassa: I think from discussions with @rbgirshick we might want to add everything model-related to the `_C.MODEL` namespace.
Could you rename this to
```python
_C.MODEL.RETINANET = CN()
```
?

maskrcnn_benchmark/modeling/backbone/fpn.py
```diff
@@ -24,9 +24,14 @@ def __init__(self, in_channels_list, out_channels, top_blocks=None):
         super(FPN, self).__init__()
         self.inner_blocks = []
         self.layer_blocks = []
+        # If in_channels is 0, it would be used. 
+        self.valid_layers = [i > 0 for i in in_channels_list]
```
fmassa: looks like you are not using this?

maskrcnn_benchmark/modeling/rpn/anchor_generator.py
```diff
@@ -54,8 +54,13 @@ def __init__(
         else:
             if len(anchor_strides) != len(sizes):
                 raise RuntimeError("FPN should have #anchor_strides == #sizes")
+
             cell_anchors = [
-                generate_anchors(anchor_stride, (size,), aspect_ratios).float()
+                generate_anchors(
+                    anchor_stride,
+                    size if type(size) is tuple else (size,),
```
fmassa: nit: can you replace this with
```python
size if isinstance(size, (tuple, list)) else (size,)
```

maskrcnn_benchmark/modeling/rpn/retinanet.py
```diff
@@ -0,0 +1,163 @@
...
+                  self.bbox_pred]:
+            for l in modules.modules():
+                if isinstance(l, nn.Conv2d):
+                    torch.nn.init.normal_(l.weight, std=0.01)
+                    torch.nn.init.constant_(l.bias, 0)
+
+
+        # retinanet_bias_init
+        prior_prob = cfg.RETINANET.PRIOR_PROB
+        bias_value = -np.log((1 - prior_prob) / prior_prob)
```
fmassa: nit: can you replace this with `math.log`?

maskrcnn_benchmark/modeling/rpn/retinanet.py
```diff
@@ -0,0 +1,163 @@
...
+        losses = {
+            "loss_retina_cls": loss_box_cls,
+            "loss_retina_reg": loss_box_reg,
+        }
+        return anchors, losses
+
+    def _forward_test(self, anchors, box_cls, box_regression):
+        boxes = self.box_selector_test(anchors, box_cls, box_regression)
+        '''
+        if self.cfg.MODEL.RPN_ONLY:
```
fmassa: Should we remove this comment?

maskrcnn_benchmark/structures/boxlist_ops.py
```diff
@@ -108,7 +108,12 @@ def cat_boxlist(bboxes):
         bboxes (list[BoxList])
     """
     assert isinstance(bboxes, (list, tuple))
-    assert all(isinstance(bbox, BoxList) for bbox in bboxes)
+
+    # bbox could be empty(list).
+    assert all(isinstance(bbox, (list, BoxList)) for bbox in bboxes)
```
fmassa: I'm not sure we want to accept lists as well.
We can have an empty `BoxList` now, as we have support for tensors with zero in its dimensions.

maskrcnn_benchmark/modeling/rpn/retinanet.py
chengyangfu: Fixed.

maskrcnn_benchmark/structures/boxlist_ops.py
chengyangfu: fixed. Use empty BoxList instead of [] in retinanet_infer.py

chengyangfu:
@rbgirshick 
   The only difference is the matching. I didn't use any anchor bbox whose IOU is between [0.4,0.5].  <br>
    For the C++ implementation of sigmoid focal loss, I have tested this in my another project. Python version needs to discard ignored examples first and then calculate focal loss, but C++/CUDA version can combine focal loss and selection. The C++/CUDA  version definitely has lower foot prints and runs faster. The critial part is the selection. If I called labels >=0 and use it to get the positive and negative exampels, the performance will drop quickly. Due to the large number of postive and negative examples in the training, I think that's reasonable. 

 The following is the python version of Focal Loss I tested. 

 ```python
    def forward(self, inputs, targets):
        N = inputs.size(0)
        C = inputs.size(1)
        class_mask = inputs.new_zeros((N, C))
        ids = targets.view(-1, 1)
        class_mask.scatter_(1, ids, 1.)

        class_mask = class_mask[:, 1:]
        inputs = inputs[:, 1:]

        P = torch.sigmoid(inputs)
        PC = P*class_mask + (1-P)*(1-class_mask)
        alpha = self.alpha * class_mask + (1 - self.alpha) * (1 - class_mask)
        focal_weight = alpha * (1 - PC).pow(self.gamma)
        loss = F.binary_cross_entropy_with_logits(inputs, class_mask,
                                                      focal_weight)
        return loss
```

**Comment by rbgirshick** [comment]
@chengyangfu thanks for the explanations!

**Comment by rbgirshick** [comment]
This PR looks good to me, though I would like to see one organizational change. I think the retinanet code that's currently inside the `rpn` package should go under a new package at `modeling/retinanet/`. @fmassa what do you think?

**Comment by fmassa** [comment]
I agree that having retinanet files be in their own folder would be preferable.
If we assume that we are keeping the GeneralizedRCNN framework for RetinaNet, and we use the retinanet heads in the `self.rpn` field, then it might make sense to split the organization as follows:
```
modeling/
  rpn/
    some_name_class_agnostic_ie_RPN/
        ... rpn files here
    retinanet/
        ... retinanet files here
    make_rpn.py # or something like that, which is the entry point in GeneralizedRCNN
```

Also, before merging this PR I'd fix a few other things as well. For example, all the other config files are currently broken because we have changed `INPUT.MIN_SIZE` to be a tuple instead of an int. That should be fixed.
There were also a few other things that I'd maybe like see changed, but I didn't yet have the time to do a full review in detail.
Maybe we could create a tag for the current version, like v0.1, and then merge this PR, and iterate a bit more on the design later on?

**Comment by chengyangfu** [comment]
According to @fmassa  and @rbgirshick opinions, all retinanet related files are moved to ``` modeling/rpn/retinanet/ ```.  
I also add a script for training RetinaNet with X_101_32x8d.

**Comment by fmassa** [comment]
@chengyangfu I'll be pulling your PR and testing it locally starting from next week.

Sorry for the wait.

**Comment by chengyangfu** [comment]
@fmassa 
Thanks for letting me know! I will rebase the branch this weekend.

**Comment by chengyangfu** [comment]
@fmassa, Rebase the code according to the new commits in the master branch. 

@rbgirshick Clarify the matching part. It is opposite to what I said.  For each gt object, the best matching anchor is always used as the positive examples even the overlapping threshold is really low.  Detectron's version discards the best matching case if the threshold is lower than 0.4.  The following table shows the performance of RetinaNet_R-50-FPN_1x with different thresholds used for the best matching case. 

| threshold | AP | AP50 | AP75 | P.S |
| --- | --- | --- | --- | --- |
| 0.5 | 35.5 | 53.7 | 38.1 | |
| 0.4 | 36.0 | 54.1 | 38.6 | Detectron version. |
| 0.3 | 36.1 | 54.5 | 38.9 | |
| 0.2 | 36.1 | 54.5 | 38.7 | | 
| 0.0 | 36.2 | 55.0 | 38.7 | This branch. |

**Comment by Z-Zheng** [comment]
Hi,
I note that the the nomalizer of smooth L1 loss is num_fg*4 in your implementation.
But I find this normalizer in Detectron version seems to be num_fg.
Which one is exact implementation?
[select_smooth_l1_loss kernel](https://github.com/pytorch/pytorch/blob/9a281451ed7f00d052212695ea112ebe872776f6/modules/detectron/select_smooth_l1_loss_op.cu#L35)
[select_smooth_l1_loss op](https://github.com/pytorch/pytorch/blob/9a281451ed7f00d052212695ea112ebe872776f6/modules/detectron/select_smooth_l1_loss_op.cu#L136)

**Comment by fmassa** [comment]
Also, @chengyangfu could you expand a bit on the different accuracies you reported?
In the main post you mentioned that there was ~0.5 improvement over Detectron, was that an oversight on the Detectron accuracies?

**Comment by chengyangfu** [comment]
> Hi,
> I note that the the nomalizer of smooth L1 loss is num_fg*4 in your implementation.
> But I find this normalizer in Detectron version seems to be num_fg.
> Which one is exact implementation?
> [select_smooth_l1_loss kernel](https://github.com/pytorch/pytorch/blob/9a281451ed7f00d052212695ea112ebe872776f6/modules/detectron/select_smooth_l1_loss_op.cu#L35)
> [select_smooth_l1_loss op](https://github.com/pytorch/pytorch/blob/9a281451ed7f00d052212695ea112ebe872776f6/modules/detectron/select_smooth_l1_loss_op.cu#L136)

Hi @Z-Zheng, 
      In Detectron, num_fg is used as the normalizer for the smooth l1 loss. But if you check the training iteration in this branch, using num_fg*4 makes the Cls and Regression losses are more balanced.

**Comment by stanstarks** [comment]
Hi, awesome work!
Training with single GPU works perfectly. 

Strange thing happens when I use torch.distributed.launch for 4GPU training as specified in the scripts, all processes halt at computing the loss in the second iteration.
Volatile GPU-Util for all devices stay at 100%.

Could you help me identify the issue? Thanks!

**Comment by chengyangfu** [comment]
> Hi, awesome work!
> Training with single GPU works perfectly.
> 
> Strange thing happens when I use torch.distributed.launch for 4GPU training as specified in the scripts, all processes halt at computing the loss in the second iteration.
> Volatile GPU-Util for all devices stay at 100%.
> 
> Could you help me identify the issue? Thanks!

Hi @stanstarks, 
   I also met this bug. You can find more information in this [thread](https://github.com/facebookresearch/maskrcnn-benchmark/issues/58).

**Comment by fmassa** [comment]
Hi @chengyangfu,

I'm pulling your PR now and I'd like to make some small modifications.
Could you give me access to push to this branch?
There is some information in https://help.github.com/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork/

**Comment by chengyangfu** [comment]
Hi @fmassa , 
     Sure. I already turned it on.

**Comment by fmassa** [comment]
Yes, thanks!

It looks like I can push here. I have a few other things I'd like to try changing.

Let me know what you'll think.

**Comment by fmassa** [comment]
One thing I'll probably be doing will be to replace the dedicated CUDA implementation of Focal Loss by a python implementation.

Thanks to the awesome work of @t-vi on the JIT side, it is possible to obtain a single optimized CUDA kernel which should be as efficient as the one from Detectron, but by writing it only in Python.

**Comment by chengyangfu** [comment]
@fmassa 
Sure. It sounds great! I will have more free time this weekend and will them carefully. Thanks for letting me know.

File: maskrcnn_benchmark/modeling/rpn/retinanet/inference.py [hunk 7]
```diff
@@ -0,0 +1,225 @@
...
+        for batch_idx, (per_box_cls, per_box_regression, per_pre_nms_top_n, \
+        per_candidate_inds, per_anchors) in enumerate(zip(
+            box_cls,
+            box_regression,
+            pre_nms_top_n,
+            candidate_inds,
+            anchors)):
+
+            # Sort and select TopN
+            per_box_cls = per_box_cls[per_candidate_inds]
```
fmassa: Can't the whole implementation that is in here accept empty tensors?
In this case, do we really need to special-case if there are no detection, as in line 72 here?

maskrcnn_benchmark/config/defaults.py
```diff
@@ -37,7 +38,7 @@
 # -----------------------------------------------------------------------------
 _C.INPUT = CN()
 # Size of the smallest side of the image during training
-_C.INPUT.MIN_SIZE_TRAIN = 800  # (800,)
+_C.INPUT.MIN_SIZE_TRAIN = (800,) # 800
```
fmassa: We are currently not using this I believe, and it conflicts with the changes in the Keypoints.
Maybe I'll just revert this part for now, or add it as a separate PR, are you ok with that?

maskrcnn_benchmark/config/defaults.py
```diff
@@ -223,6 +224,64 @@
 _C.MODEL.RESNETS.RES2_OUT_CHANNELS = 256
 _C.MODEL.RESNETS.STEM_OUT_CHANNELS = 64
 
+
+# ---------------------------------------------------------------------------- #
+# RetinaNet Options (Follow the Detectron version)
+# ---------------------------------------------------------------------------- #
+_C.MODEL.RETINANET = CN()
+
+# This is the number of foreground classes, background is not included.
```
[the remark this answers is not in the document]
chengyangfu: Yes, the comment is wrong.  For the second question, definitely, one place is much better. But I would like it is under MODEL instead of MODEL.ROI_BOX_HEAD.NUM_CLASSES.

maskrcnn_benchmark/config/defaults.py
chengyangfu: I am totally okay with that.

maskrcnn_benchmark/modeling/detector/generalized_rcnn.py
```diff
@@ -27,8 +28,12 @@ def __init__(self, cfg):
         super(GeneralizedRCNN, self).__init__()
 
         self.backbone = build_backbone(cfg)
-        self.rpn = build_rpn(cfg)
-        self.roi_heads = build_roi_heads(cfg)
+        if cfg.MODEL.RETINANET_ON:
+            self.rpn = build_retinanet(cfg)
```
fmassa: I'm thinking about moving this logic into `build_rpn`, and the same for `build_roi_heads`, so that this part keeps the same.
What do you think?

maskrcnn_benchmark/config/defaults.py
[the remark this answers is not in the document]
fmassa: Oh yes, `MODEL.ROI_BOX_HEAD.NUM_CLASSES` is not a good place to take RetinaNet into account. Let me think about it.

Also, I haven't looked closely at the retinanet implementation of Detectron, but don't we have a `background` class in the classifier?

fmassa:
@chengyangfu I did some minor modifications in the code, and I also made a few comments in places where I think we could potentially simplify the implementation by extracting a few base functions and reusing them for both RPN/Faster R-CNN and RetinaNet.
Let me know what you think

maskrcnn_benchmark/modeling/rpn/retinanet/inference.py
chengyangfu: I am not sure about this part. This version definitely is cleaner. But the early return is important to the speed performance. 
Test : RetinaNet - Resnet50 on 1080 TI) 
Previous version : 123.9 ms
This version : 149.4 ms

chengyangfu:
@fmassa Thanks! I also put some reviews there. I also like the idea of extracting base functions. Especially, the single-stage detectors(YOLO, SSD, RetinaNet) can be considered as the variant of the RPN model.

maskrcnn_benchmark/modeling/rpn/retinanet/inference.py
fmassa: That is surprising.
All the operations that happen in this for loop are on empty tensors, so there should be just a skip in the C++ side that would happen.
I'd like to understand where this difference comes from: isn't it just because of the sync point in line 94 due to the `.item()`?

Also, how often do we have images with no detections? For a trained model, that shouldn't happen often, should it?

maskrcnn_benchmark/config/defaults.py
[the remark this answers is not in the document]
fmassa: I'd rather keep the `ROI_BOX_HEAD.NUM_CLASSES` for now, as when it was defined it was a property of `ROI_HEAD`, and different heads can have different number of classes.

Thanks for the explanation on the second question!

fmassa:
Let me know if you'd like to try out factoring the base functions out (and if you'd have the time to do it).
Maybe you could start with a quick description, and we could iterate on that first?

maskrcnn_benchmark/modeling/rpn/retinanet/inference.py
chengyangfu: According to my experience, there are 2~3 images in COCO **minival** could cause the problems. (no detections with confidence >= 0.05)  Besides, I think no detection should be the normal case in inference. 
This problem also happens in Detectron. The Detectron solution is to use threshold 0.0 in the last year(P7). I also test this solution but it slows down more. (around 160~170 ms)

fmassa: So, if I understand it correctly, there are (approximately) two images over 5000 on coco for which the speed would go from 123ms to 149ms, is that right? That doesn't seem like a problem if we consider the whole evaluation time, is that right?

chengyangfu: No. Maybe I didn't express correctly.  In RetinaNet, there are 5 layers(P3-P7) used for prediction. It's very common that no detections with confidence more than 0.05 in some layers. I think that's the reason why early return can reduce the inference time. In some rare cases( 2 images over 5000 on COCO), there is no detection >0.05 in all prediction layers.

maskrcnn_benchmark/config/defaults.py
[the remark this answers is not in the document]
chengyangfu: I think that makes sense.

chengyangfu:
@fmassa  Sure. I can do it. Just let me know which part you would like to change.

maskrcnn_benchmark/modeling/rpn/retinanet/inference.py
fmassa: Oh, now I got it, thanks!
Still, I'm quite surprised by this slowdown. I'd like to understand where it comes from.
It could be due to advanced indexing, but it has been recently optimized in PyTorch master.

fmassa:
@chengyangfu I've left a few comments in the code with a `# TODO` and some quick notes saying that this part is very similar to that part. Could you have a look and see if you agree with the comments? Then we can start to plan something out of it.

maskrcnn_benchmark/modeling/rpn/retinanet/inference.py
fmassa: I think that we can obtain a speed improvement here without much effort:
we move
```python
per_candidate_nonzeros = per_candidate_inds.nonzero()
```
before this line, and pass `per_candidate_nonzeros` instead 
```python
per_box_cls = per_box_cls[per_box_loc, per_class]
```
How does this change the runtime, could you check?

chengyangfu: Okay. This step does not change current inference time much. The inference time is around 0.1477 second/per image.

t-vi: It's a bit manual, but maybe you could save a boolean flag for `is_empty = candidate_inds.sum().item() == 0` and then skip only parts of the for loop to see what exactly is causing the slowdown (i.e. does it happen when you  call the `self.box_code.decode`, does it happen if you make everything below `if per_candidate_inds_sum().item() > ...` only execute if you have `not is_empty`. That way it would be easier to pinpoint which bits the network (wastefully) spends its time in.
I imagine lots of syncing when looking at the indexing (and the .item() anyways), so I wonder whether it might even help to do the entire indexing on CPU and transfer the tensors to CPU before the for loop.

fmassa:
@chengyangfu cool, I'll try to come up with a plan tomorrow for how we could restructure those functions

zimenglan-sysu-512:
hi @fmassa 
do u have a plan to merge this pr?

fmassa:
@zimenglan-sysu-512 yes, I plan to merge this PR.

I just didn't have the time yet to propose some more concrete action plan for the refactoring that I discussed before.

fmassa:
I don't know yet. Hopefully soon, but I've been busy with other things.

configs/retina/retinanet_R-101-FPN_1x.yaml
```diff
@@ -0,0 +1,48 @@
+MODEL:
+  META_ARCHITECTURE: "GeneralizedRCNN"
+  WEIGHT: "catalog://ImageNetPretrained/MSRA/R-101"
+  RPN_ONLY: True
```
lbin: I think “PRN_ONLY” should be False, due to this option maybe affect the eval process.

tpys:
@chengyangfu 
Hi,
I fond the regression loss is so different between maskrcnn_benchmark retinanet and detectron retinanet:0.68 vs 0.25(detectron). I wonder what cause this problem,  your regression loss is twice than detectron's,  consider divide by fg_num * 4, which means 8 times than detectron's.

chengyangfu:
> @chengyangfu
> Hi,
> I find the regression loss in your branch is significantly larger (2X4) than that in the detectron. Can you elaborate more about it?

Hi @tpys , 
     I did add the normalization term(divide by 4) in [regression loss](
https://github.com/chengyangfu/maskrcnn-benchmark/blob/retinanet/maskrcnn_benchmark/modeling/rpn/retinanet/loss.py#L145). The point is to make regression and classification loss are similar during training. I didn't check the details of this part in Detectron and hope this answer your question.

configs/retina/retinanet_R-101-FPN_1x.yaml
chengyangfu: Hi, I think this does not affect anything in the evaluation. RPN_ONLY is not used, once the RETINANET_ON is set as True.

tpys:
@chengyangfu 
I see, Thanks.

configs/retina/retinanet_R-101-FPN_1x.yaml
lbin: I'm really got some errors.

Problem maybe occur on here:
https://github.com/facebookresearch/maskrcnn-benchmark/blob/0dfac376668a6618ed17c2d890c92728bb2926b6/maskrcnn_benchmark/data/datasets/evaluation/coco/coco_eval.py#L24

chengyangfu: Thanks for letting me know. I am on vacation now but will check it once I am back.

jordanxlj:
I use your patch to training coco 2017, but the loss is nan after several hundreds iterations.
2019-01-08 02:55:17,618 maskrcnn_benchmark.trainer INFO: eta: 23:50:00  iter: 180  loss: 1.3047 (1.4431)  loss_retina_cls: 1.3047 (1.4430)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4415 (0.4771)  data: 0.0054 (0.0099)  lr: 0.005720  max mem: 7676
2019-01-08 02:55:26,577 maskrcnn_benchmark.trainer INFO: eta: 23:41:05  iter: 200  loss: 1.2345 (1.4224)  loss_retina_cls: 1.2345 (1.4224)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4520 (0.4742)  data: 0.0051 (0.0095)  lr: 0.005987  max mem: 7676
2019-01-08 02:55:35,463 maskrcnn_benchmark.trainer INFO: eta: 23:32:47  iter: 220  loss: 1.2039 (1.4037)  loss_retina_cls: 1.2039 (1.4037)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4464 (0.4715)  data: 0.0047 (0.0092)  lr: 0.006253  max mem: 7676
2019-01-08 02:55:44,289 maskrcnn_benchmark.trainer INFO: eta: 23:25:05  iter: 240  loss: 1.2018 (1.3867)  loss_retina_cls: 1.2018 (1.3867)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4451 (0.4690)  data: 0.0053 (0.0090)  lr: 0.006520  max mem: 7676
2019-01-08 02:55:53,262 maskrcnn_benchmark.trainer INFO: eta: 23:20:14  iter: 260  loss: 1.1639 (1.3700)  loss_retina_cls: 1.1639 (1.3700)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4459 (0.4674)  data: 0.0049 (0.0088)  lr: 0.006787  max mem: 7676
2019-01-08 02:56:02,121 maskrcnn_benchmark.trainer INFO: eta: 23:14:51  iter: 280  loss: 1.1376 (1.3538)  loss_retina_cls: 1.1376 (1.3538)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4455 (0.4657)  data: 0.0052 (0.0086)  lr: 0.007053  max mem: 7676
2019-01-08 02:56:10,982 maskrcnn_benchmark.trainer INFO: eta: 23:10:11  iter: 300  loss: 1.1163 (1.3383)  loss_retina_cls: 1.1163 (1.3383)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4415 (0.4642)  data: 0.0060 (0.0085)  lr: 0.007320  max mem: 7676
2019-01-08 02:56:19,854 maskrcnn_benchmark.trainer INFO: eta: 23:06:10  iter: 320  loss: 1.0964 (1.3243)  loss_retina_cls: 1.0964 (1.3243)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4452 (0.4629)  data: 0.0061 (0.0084)  lr: 0.007587  max mem: 7676
2019-01-08 02:56:28,723 maskrcnn_benchmark.trainer INFO: eta: 23:02:36  iter: 340  loss: nan (nan)  loss_retina_cls: nan (nan)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4465 (0.4617)  data: 0.0056 (0.0083)  lr: 0.007853  max mem: 7676
2019-01-08 02:56:37,588 maskrcnn_benchmark.trainer INFO: eta: 22:59:22  iter: 360  loss: nan (nan)  loss_retina_cls: nan (nan)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4414 (0.4607)  data: 0.0056 (0.0082)  lr: 0.008120  max mem: 7676
2019-01-08 02:56:46,850 maskrcnn_benchmark.trainer INFO: eta: 22:59:35  iter: 380  loss: nan (nan)  loss_retina_cls: nan (nan)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4470 (0.4608)  data: 0.0061 (0.0082)  lr: 0.008387  max mem: 7676
2019-01-08 02:56:55,639 maskrcnn_benchmark.trainer INFO: eta: 22:56:14  iter: 400  loss: nan (nan)  loss_retina_cls: nan (nan)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4425 (0.4598)  data: 0.0075 (0.0081)  lr: 0.008653  max mem: 7676
2019-01-08 02:57:04,386 maskrcnn_benchmark.trainer INFO: eta: 22:52:53  iter: 420  loss: nan (nan)  loss_retina_cls: nan (nan)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4361 (0.4587)  data: 0.0062 (0.0080)  lr: 0.008920  max mem: 7676
2019-01-08 02:57:13,233 maskrcnn_benchmark.trainer INFO: eta: 22:50:30  iter: 440  loss: nan (nan)  loss_retina_cls: nan (nan)  loss_retina_reg: 0.0000 (0.0000)  time: 0.4418 (0.4580)  data: 0.0057 (0.0080)  lr: 0.009187  max mem: 7676

chengyangfu:
Hi @jordanxlj  , 
    I will update the code according to the new things in master branch in this week and will also check this problem. But at the same time, could you provide more information? For example, GPU, CUDA, and which config you are using. Besides, did you change any parameter, such as batch size or learning rate ?

Z-Zheng:
> I use your patch to training coco 2017, the model is R-50-FPN, lr is 0.01, but the loss is nan after several hundreds iterations.
> 2019-01-08 02:55:17,618 maskrcnn_benchmark.trainer INFO: eta: 23:50:00 iter: 180 loss: 1.3047 (1.4431) loss_retina_cls: 1.3047 (1.4430) loss_retina_reg: 0.0000 (0.0000) time: 0.4415 (0.4771) data: 0.0054 (0.0099) lr: 0.005720 max mem: 7676
> 2019-01-08 02:55:26,577 maskrcnn_benchmark.trainer INFO: eta: 23:41:05 iter: 200 loss: 1.2345 (1.4224) loss_retina_cls: 1.2345 (1.4224) loss_retina_reg: 0.0000 (0.0000) time: 0.4520 (0.4742) data: 0.0051 (0.0095) lr: 0.005987 max mem: 7676
> 2019-01-08 02:55:35,463 maskrcnn_benchmark.trainer INFO: eta: 23:32:47 iter: 220 loss: 1.2039 (1.4037) loss_retina_cls: 1.2039 (1.4037) loss_retina_reg: 0.0000 (0.0000) time: 0.4464 (0.4715) data: 0.0047 (0.0092) lr: 0.006253 max mem: 7676
> 2019-01-08 02:55:44,289 maskrcnn_benchmark.trainer INFO: eta: 23:25:05 iter: 240 loss: 1.2018 (1.3867) loss_retina_cls: 1.2018 (1.3867) loss_retina_reg: 0.0000 (0.0000) time: 0.4451 (0.4690) data: 0.0053 (0.0090) lr: 0.006520 max mem: 7676
> 2019-01-08 02:55:53,262 maskrcnn_benchmark.trainer INFO: eta: 23:20:14 iter: 260 loss: 1.1639 (1.3700) loss_retina_cls: 1.1639 (1.3700) loss_retina_reg: 0.0000 (0.0000) time: 0.4459 (0.4674) data: 0.0049 (0.0088) lr: 0.006787 max mem: 7676
> 2019-01-08 02:56:02,121 maskrcnn_benchmark.trainer INFO: eta: 23:14:51 iter: 280 loss: 1.1376 (1.3538) loss_retina_cls: 1.1376 (1.3538) loss_retina_reg: 0.0000 (0.0000) time: 0.4455 (0.4657) data: 0.0052 (0.0086) lr: 0.007053 max mem: 7676
> 2019-01-08 02:56:10,982 maskrcnn_benchmark.trainer INFO: eta: 23:10:11 iter: 300 loss: 1.1163 (1.3383) loss_retina_cls: 1.1163 (1.3383) loss_retina_reg: 0.0000 (0.0000) time: 0.4415 (0.4642) data: 0.0060 (0.0085) lr: 0.007320 max mem: 7676
> 2019-01-08 02:56:19,854 maskrcnn_benchmark.trainer INFO: eta: 23:06:10 iter: 320 loss: 1.0964 (1.3243) loss_retina_cls: 1.0964 (1.3243) loss_retina_reg: 0.0000 (0.0000) time: 0.4452 (0.4629) data: 0.0061 (0.0084) lr: 0.007587 max mem: 7676
> 2019-01-08 02:56:28,723 maskrcnn_benchmark.trainer INFO: eta: 23:02:36 iter: 340 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4465 (0.4617) data: 0.0056 (0.0083) lr: 0.007853 max mem: 7676
> 2019-01-08 02:56:37,588 maskrcnn_benchmark.trainer INFO: eta: 22:59:22 iter: 360 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4414 (0.4607) data: 0.0056 (0.0082) lr: 0.008120 max mem: 7676
> 2019-01-08 02:56:46,850 maskrcnn_benchmark.trainer INFO: eta: 22:59:35 iter: 380 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4470 (0.4608) data: 0.0061 (0.0082) lr: 0.008387 max mem: 7676
> 2019-01-08 02:56:55,639 maskrcnn_benchmark.trainer INFO: eta: 22:56:14 iter: 400 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4425 (0.4598) data: 0.0075 (0.0081) lr: 0.008653 max mem: 7676
> 2019-01-08 02:57:04,386 maskrcnn_benchmark.trainer INFO: eta: 22:52:53 iter: 420 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4361 (0.4587) data: 0.0062 (0.0080) lr: 0.008920 max mem: 7676
> 2019-01-08 02:57:13,233 maskrcnn_benchmark.trainer INFO: eta: 22:50:30 iter: 440 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4418 (0.4580) data: 0.0057 (0.0080) lr: 0.009187 max mem: 7676

Why your loss_retina_reg is zero?

jordanxlj:
> > I use your patch to training coco 2017, the model is R-50-FPN, lr is 0.01, but the loss is nan after several hundreds iterations.
> > 2019-01-08 02:55:17,618 maskrcnn_benchmark.trainer INFO: eta: 23:50:00 iter: 180 loss: 1.3047 (1.4431) loss_retina_cls: 1.3047 (1.4430) loss_retina_reg: 0.0000 (0.0000) time: 0.4415 (0.4771) data: 0.0054 (0.0099) lr: 0.005720 max mem: 7676
> > 2019-01-08 02:55:26,577 maskrcnn_benchmark.trainer INFO: eta: 23:41:05 iter: 200 loss: 1.2345 (1.4224) loss_retina_cls: 1.2345 (1.4224) loss_retina_reg: 0.0000 (0.0000) time: 0.4520 (0.4742) data: 0.0051 (0.0095) lr: 0.005987 max mem: 7676
> > 2019-01-08 02:55:35,463 maskrcnn_benchmark.trainer INFO: eta: 23:32:47 iter: 220 loss: 1.2039 (1.4037) loss_retina_cls: 1.2039 (1.4037) loss_retina_reg: 0.0000 (0.0000) time: 0.4464 (0.4715) data: 0.0047 (0.0092) lr: 0.006253 max mem: 7676
> > 2019-01-08 02:55:44,289 maskrcnn_benchmark.trainer INFO: eta: 23:25:05 iter: 240 loss: 1.2018 (1.3867) loss_retina_cls: 1.2018 (1.3867) loss_retina_reg: 0.0000 (0.0000) time: 0.4451 (0.4690) data: 0.0053 (0.0090) lr: 0.006520 max mem: 7676
> > 2019-01-08 02:55:53,262 maskrcnn_benchmark.trainer INFO: eta: 23:20:14 iter: 260 loss: 1.1639 (1.3700) loss_retina_cls: 1.1639 (1.3700) loss_retina_reg: 0.0000 (0.0000) time: 0.4459 (0.4674) data: 0.0049 (0.0088) lr: 0.006787 max mem: 7676
> > 2019-01-08 02:56:02,121 maskrcnn_benchmark.trainer INFO: eta: 23:14:51 iter: 280 loss: 1.1376 (1.3538) loss_retina_cls: 1.1376 (1.3538) loss_retina_reg: 0.0000 (0.0000) time: 0.4455 (0.4657) data: 0.0052 (0.0086) lr: 0.007053 max mem: 7676
> > 2019-01-08 02:56:10,982 maskrcnn_benchmark.trainer INFO: eta: 23:10:11 iter: 300 loss: 1.1163 (1.3383) loss_retina_cls: 1.1163 (1.3383) loss_retina_reg: 0.0000 (0.0000) time: 0.4415 (0.4642) data: 0.0060 (0.0085) lr: 0.007320 max mem: 7676
> > 2019-01-08 02:56:19,854 maskrcnn_benchmark.trainer INFO: eta: 23:06:10 iter: 320 loss: 1.0964 (1.3243) loss_retina_cls: 1.0964 (1.3243) loss_retina_reg: 0.0000 (0.0000) time: 0.4452 (0.4629) data: 0.0061 (0.0084) lr: 0.007587 max mem: 7676
> > 2019-01-08 02:56:28,723 maskrcnn_benchmark.trainer INFO: eta: 23:02:36 iter: 340 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4465 (0.4617) data: 0.0056 (0.0083) lr: 0.007853 max mem: 7676
> > 2019-01-08 02:56:37,588 maskrcnn_benchmark.trainer INFO: eta: 22:59:22 iter: 360 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4414 (0.4607) data: 0.0056 (0.0082) lr: 0.008120 max mem: 7676
> > 2019-01-08 02:56:46,850 maskrcnn_benchmark.trainer INFO: eta: 22:59:35 iter: 380 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4470 (0.4608) data: 0.0061 (0.0082) lr: 0.008387 max mem: 7676
> > 2019-01-08 02:56:55,639 maskrcnn_benchmark.trainer INFO: eta: 22:56:14 iter: 400 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4425 (0.4598) data: 0.0075 (0.0081) lr: 0.008653 max mem: 7676
> > 2019-01-08 02:57:04,386 maskrcnn_benchmark.trainer INFO: eta: 22:52:53 iter: 420 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4361 (0.4587) data: 0.0062 (0.0080) lr: 0.008920 max mem: 7676
> > 2019-01-08 02:57:13,233 maskrcnn_benchmark.trainer INFO: eta: 22:50:30 iter: 440 loss: nan (nan) loss_retina_cls: nan (nan) loss_retina_reg: 0.0000 (0.0000) time: 0.4418 (0.4580) data: 0.0057 (0.0080) lr: 0.009187 max mem: 7676
> 
> Why your loss_retina_reg is zero?

I modify the loss computation, I found the bug and fixed it, but it is still nan after hundreds iterations.

jordanxlj:
> Hi @jordanxlj ,
> I will update the code according to the new things in master branch in this week and will also check this problem. But at the same time, could you provide more information? For example, GPU, CUDA, and which config you are using. Besides, did you change any parameter, such as batch size or learning rate ?

I use 8 GPU V100, batch size is 8, learning rate is 0.01, I modify the learning rate into 0.001 or 0.0001, the iteration would be larger, but nan still will occur.

jordanxlj:
> > Hi @jordanxlj ,
> > I will update the code according to the new things in master branch in this week and will also check this problem. But at the same time, could you provide more information? For example, GPU, CUDA, and which config you are using. Besides, did you change any parameter, such as batch size or learning rate ?
> 
> I use 8 GPU V100, batch size is 8, learning rate is 0.01, I modify the learning rate into 0.001 or 0.0001, the iteration would be larger, but nan still will occur.

I suspected the computation of sigmoid focal loss, so I added prob max and min value clamp, but it's not usable.

zimenglan-sysu-512:
hi @jordanxlj 
i also meet this problem,  i change these [lines](https://github.com/facebookresearch/maskrcnn-benchmark/blob/8bab2389ce/maskrcnn_benchmark/modeling/rpn/retinanet/loss.py#L140-L153) as below to solve it:
```
        if pos_inds.shape[0] > 0:
            reduction = self.reduction
            retinanet_regression_loss = smooth_l1_loss(
                box_regression[pos_inds],
                regression_targets[pos_inds],
                beta=self.bbox_reg_beta,
                reduction=reduction,
            ) / (pos_inds.numel() * 4) * self.bbox_reg_weight
        if pos_inds.shape[0] == 0 or torch.isnan(retinanet_regression_loss):
            # print("\nretinanet_regression_loss is set to zero...\n")
            retinanet_regression_loss = (0.0 * box_regression).sum()

        labels = labels.int()

        if pos_inds.shape[0] > 0:
            retinanet_cls_loss = self.box_cls_loss_func(
                box_cls,
                labels
            ) / (pos_inds.numel() + N)  # TODO why the N?
        if pos_inds.shape[0] == 0 or torch.isnan(retinanet_cls_loss):
            # print("\nretinanet_cls_loss is set to zero...\n")
            retinanet_cls_loss = (0.0 * box_cls).sum()
```

jordanxlj:
> hi @jordanxlj
> i also meet this problem, i change these [lines](https://github.com/facebookresearch/maskrcnn-benchmark/blob/8bab2389ce/maskrcnn_benchmark/modeling/rpn/retinanet/loss.py#L140-L153) as below to solve it:
> 
> ```
>         if pos_inds.shape[0] > 0:
>             reduction = self.reduction
>             retinanet_regression_loss = smooth_l1_loss(
>                 box_regression[pos_inds],
>                 regression_targets[pos_inds],
>                 beta=self.bbox_reg_beta,
>                 reduction=reduction,
>             ) / (pos_inds.numel() * 4) * self.bbox_reg_weight
>         if pos_inds.shape[0] == 0 or torch.isnan(retinanet_regression_loss):
>             # print("\nretinanet_regression_loss is set to zero...\n")
>             retinanet_regression_loss = (0.0 * box_regression).sum()
> 
>         labels = labels.int()
> 
>         if pos_inds.shape[0] > 0:
>             retinanet_cls_loss = self.box_cls_loss_func(
>                 box_cls,
>                 labels
>             ) / (pos_inds.numel() + N)  # TODO why the N?
>         if pos_inds.shape[0] == 0 or torch.isnan(retinanet_cls_loss):
>             # print("\nretinanet_cls_loss is set to zero...\n")
>             retinanet_cls_loss = (0.0 * box_cls).sum()
> ```

Thanks, I have already tried adding group norm to the FPN P6/P7 and the RetinanetHeads, but it didn't take effect. Should the gn layer's group number not be appropriate?

beaulian:
Hi @chengyangfu ! Before I saw your implementation of RetinaNet, I implemented a version of RetinaNet by myself. So I have some questions: 

- In your class `LastLevelP6P7`, why did you take as P5 the input of P6 instead of C5?
- can I not divide by 4 in regression loss?  does the additional operation(divide by 4) not hurt the results?

Thanks! Looking forward to your reply!

zimenglan-sysu-512:
hi @jordanxlj 
i pr #346 , maybe u can have a look

chengyangfu:
Hi @beaulian, 
      For the question1, you can check the footnote2 in page4 in focal loss paper. 
      Feel free to remove the "divide by 4 part" and test it by yourself. It just makes the losses of classification and regression more balanced. If you train the network long enough, I do not think it will affect the performance. 


Focalloss: https://arxiv.org/pdf/1708.02002.pdf

beaulian:
Hi @chengyangfu ,
Thank you for reply! I have seen the paper. However, the paper say that 'P6 is obtained via a 3×3 stride-2 conv on C5'. The reason why I ask you for the `LastLevelP6P7` is that the paper use C5 instead of P5 for the P6.

Thanks! Looking forward to your reply!

chengyangfu:
Hi @beaulian, 
      Yes, you are right. Original paper uses C5 to generates P6 but I use P5 in this repository. I will train the model using C5 later, maybe this weekend to check the accuracy. But I do not think this will affect the accuracy much.

fmassa:
@chengyangfu thanks, please let us know once you finish your trainings.
I'll be looking into finally work on merging this PR in the next week or so

chengyangfu:
@fmassa  Sure. Sorry for the delay I was too busy and traveled a lot in the last month. Will have more time to work on this after this Friday.

fmassa:
@chengyangfu sure, no worries!
Also, great work with RetinaMask! Would you mind sending a PR creating a new section in the README `Projects using maskrcnn-benchmark` and link to your project / paper?

chengyangfu:
@fmassa Sure! will do it this weekend.

beaulian:
@chengyangfu Thank you for the reply! There still exists a question. I cannot reimplement your mAP.
Here is my result:
`Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.238
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.372
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.252
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.111
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.248
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.323
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.263
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.424
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.442
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.240
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.470
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.603
2019-01-16 21:21:01,212 maskrcnn_benchmark.inference INFO: OrderedDict([('bbox', OrderedDict([('AP', 0.23835689408090532), ('AP50', 0.37218548504960325), ('AP75', 0.2517431195117994), ('APs', 0.11133113085067697), ('APm', 0.24752989953418778), ('APl', 0.3230093281236464)]))])` 

The network is `retinanet_R-50-FPN_1x`. For the parameters, the `IMS_PER_BATCH` is 8 for 4 GPUs, others are the same as yours. 

Thanks! Looking forward to your reply!

chengyangfu:
Hi @beaulian, 
 That mAP is too low. When you set the  `IMS_PER_BATCH`  as 8, did you follow the Linear Scaling policy to change the learning rate and the number of iterations at the same time? Your config should look like the following. 

Original (batch size 16) 
`BASE_LR`: 0.01
`STEPS`: (60000, 80000)
`MAX_ITER`: 90000
`IMS_PER_BATCH`: 16

Modified (batch size 8)
`BASE_LR`: 0.005
`STEPS`: (120000, 160000)
`MAX_ITER`: 180000
`IMS_PER_BATCH`: 8

beaulian:
@chengyangfu Thanks! I will try again.

beaulian:
Hi, @chengyangfu 
By removing the "divide by 4" and using the following parameters:

> `BASE_LR`: 0.005
`STEPS`: (120000, 160000)
`MAX_ITER`: 180000
`IMS_PER_BATCH`: 8

I get the following result (use 4 GPUs):

> Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.317
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.480
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.336
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.158
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.335
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.431
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.300
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.481
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.503
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.294
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.542
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.660
2019-01-18 10:27:35,365 maskrcnn_benchmark.inference INFO: OrderedDict([('bbox', OrderedDict([('AP', 0.3171265512812674), ('AP50', 0.47972454402463255), ('AP75', 0.33577295081551967), ('APs', 0.15846158317045006), ('APm', 0.3353392627555818), ('APl', 0.4306224164403989)]))])

beaulian:
Hi @chengyangfu ,
Ok, I will try the "divide by 4" once I get the HPC resource from my school.

chengyangfu:
Hi @beaulian, 
    I trained a model without using "divide by 4" in the regression loss and, got similar results. I add the results at the beginning of this post. 
    I do not have a scientific way to explain why using "divide by 4" is better but according to my understanding, keeping the losses with similar magnitudes should give the better accuracy. That's why I add the "divide by 4".  
    There are some research works discussing how to combine the losses in multi-tasking. You can check them. I put one of them in the following. 
[Multi-Task Learning Using Uncertainty to Weigh Losses for Scene Geometry and Semantics, CVPR 2018](https://arxiv.org/abs/1705.07115)

chengyangfu:
Hi @beaulian, 
   Glad it works. I think the `mAP` is good and the difference is in the reasonable range.

chengyangfu:
> Hi @chengyangfu ,
> Thank you for reply! I have seen the paper. However, the paper say that `P6 is obtained via a 3×3 stride-2 conv on C5`. The reason why I ask you for the `LastLevelP6P7` is that the paper use C5 instead of P5 for the P6.
> 
> Thanks! Looking forward to your reply!

Hi @beaulian, 
Thanks for letting me know this. 
I put the results using P5 or C5 at the beginning of the post and make the defaults setting using C5. 
The performance difference is pretty minor. But I think P5 is slightly more efficient in memory and speed.

chengyangfu:
Hi @beaulian 
    I have tested using `P5` or `C5` and put the result at the beginning of the thread. The difference is really minor.  If you use more predictions for testing, definitely will get better accuracy. But your speed performance will drop significantly. 
P.S. COCO evaluates top 100 predictions per category per image. So you can just your parameter from 2000 to 100.

beaulian:
Hi @chengyangfu ,
Got it! Thanks!

fmassa:
@chengyangfu I saw that you simplified part of the implementation for the loss computation, awesome!

I'll give this PR another pass, but I think that this looks very good. One thing that is a bit annoying is that there might be some subtle differences with Detectron, given that there are some things that we need here that are not present in Detectron (I'm thinking about the +4)?

But now I saw that you changed the +4 to be the `len(box_cls)`? Is that what detectron does?

chengyangfu:
Hi @fmassa, 

I am not sure which part you mentioned. Is it the normalization parameter used in the [classifcation loss](https://github.com/facebookresearch/maskrcnn-benchmark/blob/dc73a33e3042462669bdd164c9dfd841e17e3c9a/maskrcnn_benchmark/modeling/rpn/retinanet/loss.py#L68)?  If not using the `*4` part, the performance result(29.6AP) is really low. 

I think the main difference between this and Detectron is how normalization used in the classification error(focalloss). In Detectron, the error is normalized in each layer separately. But, I sum the error first and normalize them(This is similar to the original SSD design).  
Detectron : 
<a href="https://www.codecogs.com/eqnedit.php?latex=\sum_{i\subseteq&space;layers}&space;(L_i&space;/&space;(p_i&space;&plus;&space;1))" target="_blank"></a>

Here: 
<a href="https://www.codecogs.com/eqnedit.php?latex=(\sum_{i\subseteq&space;layers}&space;L_i)&space;/&space;((\sum_{i\subseteq&space;layers}&space;p_i&space;)&space;&plus;&space;layers)" target="_blank"></a>

, where  `L_i` is the focalloss from one prediction layer and `p_i` is the number of positive examples. 

I am going to simplify the inference part first and then test the normalization of Detectron.

fmassa:
@chengyangfu oh I see now what [this part corresponds to](https://github.com/facebookresearch/maskrcnn-benchmark/blob/dc73a33e3042462669bdd164c9dfd841e17e3c9a/maskrcnn_benchmark/modeling/rpn/retinanet/loss.py#L75), thanks!

Yes, I was still a bit annoyed by from where [this `*4` should come from](https://github.com/facebookresearch/maskrcnn-benchmark/blob/dc73a33e3042462669bdd164c9dfd841e17e3c9a/maskrcnn_benchmark/modeling/rpn/retinanet/loss.py#L68). I couldn't find anything similar in Detectron, and it's weird that without it the performances are much worse.

> I am going to simplify the inference part first and then test the normalization of Detectron.

Awesome, thanks a lot!

chengyangfu:
Hi @fmassa, 
     I found out the reason why this branch giving 0.6 higher AP than Detectron and add the table to the beginning of this post. The weights of BoxCoder are different. 
In this branch, I use the (10, 10, 5, 5) in BoxCoder. Detectron adopts (1, 1, ,1, 1). That's the reason why the scales of regression loss are different and this branch needs (*4) to balance the classification and regression losses.

chengyangfu:
<a href="https://www.codecogs.com/eqnedit.php?latex=Loss&space;=&space;L_{cls}&space;&plus;&space;L_{reg}/\alpha" target="_blank"></a>

Test different <a href="https://www.codecogs.com/eqnedit.php?latex=\alpha" target="_blank"></a> in the following table. 

| Norm | AP | AP50 | AP75 | APs | APm | APl |  |
| ------  | ---- | ------  | ------  | ----- | ------ | ---- | --- |
| 1 | 31.8 | 48.1 | 34.0 | 16.3 | 33.6 | 43.1 | |
| 2 | 35.4 | 53.4 | 38.1 | 19.2 | 38.4 | 47.5 | |
| 4 | 36.2 | 55.3 | 38.6 | 20.2 | 39.8 | 48.7 | used here |
| 6 | 36.1 | 55.7 | 38.3 | 20.3 | 39.5 | 48.4 | | 
| 8 | 55.9 | 55.9 | 38.4 | 19.5 | 39.8 | 48.6 | |

fmassa:
Hi @chengyangfu 

Thanks a lot for finding out where the difference comes from! This is very interesting! cc @rbgirshick about why this PR gives better results than Detectron implementation of RetinaNet.

About your table, I was surprised when I saw the `Norm 8: AP: 55.9`, and then realized it was probably a copy-paste problem with the AP50 :-)

@chengyangfu I'll give one more pass, but I believe this is good to go on your side?

maskrcnn_benchmark/modeling/backbone/fpn.py
```diff
@@ -59,7 +64,11 @@ def forward(self, x):
             last_inner = inner_lateral + inner_top_down
             results.insert(0, getattr(self, layer_block)(last_inner))
 
-        if self.top_blocks is not None:
+        # if self.top_blocks is not None:
```
fmassa: nit: remove the comment

maskrcnn_benchmark/layers/sigmoid_focal_loss.py
```diff
@@ -0,0 +1,70 @@
+import torch
+from torch import nn
+# from torch.autograd import Function
+# from torch.autograd.function import once_differentiable
+
+# from maskrcnn_benchmark import _C
+
+"""
+class _SigmoidFocalLoss(Function):
```
fmassa: question: did your last experiments used the CUDA implementation or the Python implementation?
Could we maybe remove the CUDA code if you used the Python one? I'm thinking about using `torch.jit.script` to fuse everything into a single kernel, so we shouldn't see much performance penalty.

maskrcnn_benchmark/modeling/rpn/utils.py
```diff
@@ -0,0 +1,50 @@
...
+        )
+        box_regression_flattened.append(box_regression_per_level)
+    # concatenate on the first dimension (representing the feature levels), to
+    # take into account the way the labels were generated (with all feature maps
+    # being concatenated as well)
+    box_cls = cat(box_cls_flattened, dim=1).reshape(-1, C)
+    box_regression = cat(box_regression_flattened, dim=1).reshape(-1, 4)
+    return box_cls, box_regression
+
+
```
fmassa: nit: remove training newlines

laibe:
Hi, thanks again @chengyangfu for the great work. I've added a MobileNetV2 variation and run some benchmarks on my machine:  021b12e6d68f40f2372a3feed24ab4b04c4dd67c 

 config                                | AP    | AP_50 | AP_75 | AP_S  | AP_M  | AP_L  | 
|---------------------------------------|-------|-------|-------|-------|-------|-------|
| retinanet_R-101-FPN_1x                | 0.385 | 0.579 | 0.413 | 0.210 | 0.428 | 0.514 | 
| retinanet_R-50-FPN_1x                 | 0.362 | 0.551 | 0.388 | 0.197 | 0.396 | 0.487 |
| retinanet_MobileNetV2-FPN_1x          | 0.283 | 0.448 | 0.297 | 0.142 | 0.306 | 0.389 |
| retinanet_MobileNetV2-96-FPN_1x       | 0.251 | 0.401 | 0.264 | 0.120 | 0.275 | 0.342 |


| config                                | GPU s / img    | CPU  s / img  | Params | MAdds | Size  | AP    | 
|---------------------------------------|--------|--------|--------|-------|-------|-------|
| retinanet_R-101-FPN_1x                | 0.1858 | 2.7092 | 52.7M  | 30.9B | 402MB | 0.385 | 
| retinanet_R-50-FPN_1x                 | 0.0957 | 2.2511 | 33.8M  | 23.3B | 257MB | 0.362 | 
| retinanet_MobileNetV2-FPN_1x          | 0.0797 | 1.5275 | 12M    | 15.7B | 92MB  | 0.283 |
| retinanet_MobileNetV2-96-FPN_1x       | 0.0674 | 0.8764 | 4.1M   | 3.6B  | 31MB  | 0.251 |

retinanet_MobileNetV2-96-FPN_1x has OUT_CHANNELS: 96 in the FPN instead of 256. 
**GPU**: 1x NVIDIA GeForce GTX 1080Ti 
**CPU**: AMD Ryzen™ 7 1700 

pretrained weights: 
[retinanet_MobileNetV2-FPN_1x.pth](https://dl.dropboxusercontent.com/s/wnyjzmhhep2smjl/retinanet_MobileNetV2-FPN_1x.pth) 
[retinanet_MobileNetV2-96-FPN_1x.pth](https://dl.dropboxusercontent.com/s/qxy3n7q20cvpyq6/retinanet_MobileNetV2-96-FPN_1x.pth)

fmassa:
@laibe cool, very interesting!

Do you have the runtime estimates for those models?

Also, I suppose you have based your mobilenet implementation from https://github.com/facebookresearch/maskrcnn-benchmark/pull/242 ?

chengyangfu:
@fmassa 
   Thanks! `Norm 8: AP: 55.9`  definitely is a typo and I just updated the table again. 
    The current version is using CUDA implementation, but I will try `torch.jit.script` and test it this week.

chengyangfu:
Hi @laibe, 
  This looks good! I will check it later!

laibe:
@fmassa they are pretty similar and both based on @tonylins mobilenetv2 implementation. 
#242 uses [3, 6, 10, 18], whereas I used [3, 6, **13**, 18] as output layers. 13 is the last one with 1/16 resolution so I picked that one.

maskrcnn_benchmark/data/transforms/build.py
```diff
@@ -7,10 +7,12 @@ def build_transforms(cfg, is_train=True):
         min_size = cfg.INPUT.MIN_SIZE_TRAIN
         max_size = cfg.INPUT.MAX_SIZE_TRAIN
         flip_prob = 0.5  # cfg.INPUT.FLIP_PROB_TRAIN
+        resize = T.MultiScaleResize(min_size, max_size)
```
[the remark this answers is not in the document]
chengyangfu: k.

maskrcnn_benchmark/data/transforms/transforms.py
```diff
@@ -61,6 +61,19 @@ def __call__(self, image, target):
         return image, target
 
 
+class MultiScaleResize(object):
```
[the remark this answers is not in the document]
chengyangfu: k.

maskrcnn_benchmark/modeling/backbone/fpn.py
chengyangfu: k

maskrcnn_benchmark/modeling/rpn/utils.py
chengyangfu: k

chengyangfu:
@fmassa   
      
Experiment setting : 
PyTorch : vr.1.0.0
GPU : Pascal TitanX 
Model : retinanet_R-50-FPN_1x 
Batch size :2 for one GPU. 
`Time` and `Memory` are measured after 1000 iteration. 

| Implementation | Time | Memory |
| -------------------- | ------ | ----------- |
| CUDA | 0.3846 | 3443 | 
| Python | 0.4114 | 5298 | 
| JIT |0.4145 | 6661 |

Looks like CUDA implementation still has some advantages over time and memory. But I am surprised by the performance of `JIT` method. Do you have any insight about this?

zimenglan-sysu-512:
hi @laibe 
i try to run your yaml files, it encounter OOM (IMS_PER_BATCH=2)

fmassa:
Hi @chengyangfu ,

Thanks for the benchmark!

I believe this is a consequence of the operations not being fused in PyTorch 1.0.0. I think there have been some improvements recently that made it better, but I'd need to check.

cc @ailzhang for the performance timings and memory

Let's keep the CUDA implementation for now then, and dispatch to the Python implementation if we the tensor is on CPU, how does that sound?

Then this will be ready to be merged.

Once again, thanks a lot for this awesome contribution!

laibe:
@zimenglan-sysu-512 what's your training setup? I was using:
1x GeForce GTX 1080 Ti  
CUDA runtime version: 9.0.176 
with max mem during training of  5132

zimenglan-sysu-512:
thanks @laibe 
i make a mistake resulting OOM. 
btw, do u use CUDA implementation or Python implementation?

ailzhang:
The numbers might make sense given the current fusion logic in jit, @waochaol @zou3519 could you also help check on the JIT numbers? Thanks!

chengyangfu:
@fmassa 
    It sounds good to me. I just updated the `maskrcnn_benchmark/layers/sigmoid_focal_loss.py ` according to the suggestion.

fmassa:
That's an awesome contribution @chengyangfu , thanks a lot for all your effort!

buaaMars:
@chengyangfu 
Hi,
Great work!
I have a question on rpn/retinanet/inference.py(77)
Why do you reshape the `box_regression` when it has been `permute_and_flatten` just in the last line? According to rpn/utils.py(13), (N, -1, 4) is exactly the shape of `box_regression`. Do you reshape it again in order to handle with some spacial cases?

thks a lot

as1392:
Ah... I finally realized why model zoo does not have these trained weights...  Removing OUT_CHANNELS: 256 from backbone destroyed trained networks? I hope someone update/convert these weights :(

dedoogong:
could you please support Sigmoid Focal Loss cuda implementation to run on FP16?

Thank you

simaiden:
Can I use this model to train my custom dataset as in #521 ?

ChenJoya:
supplement performance of retinanet_r101fpn_2x on COCO minival:
AP, AP50, AP75, APs, APm, APl
0.3878, 0.5811, 0.4132, 0.2081, 0.4256, 0.5183

adizhol:
Hi @chengyangfu :)
Why is the focal loss (the sum of the losses) divided by the number of positive labels plus number of labels (N = len(labels))?
```
retinanet_cls_loss = self.box_cls_loss_func(
            box_cls,
            labels
        ) / (pos_inds.numel() + N)
```

PR #124, opened 2018-11-06
Add an option to postprocess masks during inference

hadim:
Training and inference work ok.

The only thing is that more than one image per batch is broken during inference because `Masker` does not support more than one image.

### DIFF
--- a/maskrcnn_benchmark/engine/inference.py
+++ b/maskrcnn_benchmark/engine/inference.py
@@ -91,7 +91,8 @@ def prepare_for_coco_segmentation(predictions, dataset):
         # t = time.time()
         # Masker is necessary only if masks haven't been already resized.
         if list(masks.shape[-2:]) != [image_height, image_width]:
-            masks = masker(masks, prediction)
+            masks = masker(masks.expand(1, -1, -1, -1, -1), prediction)
+            masks = masks[0]
         # logger.info('Time mask: {}'.format(time.time() - t))
         # prediction = prediction.convert('xywh')
 
--- a/maskrcnn_benchmark/engine/inference.py
+++ b/maskrcnn_benchmark/engine/inference.py
@@ -87,8 +87,11 @@ def prepare_for_coco_segmentation(predictions, dataset):
         image_height = dataset.coco.imgs[original_id]["height"]
         prediction = prediction.resize((image_width, image_height))
         masks = prediction.get_field("mask")
+        
         # t = time.time()
-        masks = masker(masks, prediction)
+        # Masker is necessary only if masks haven't been already resized.
+        if list(masks.shape[-2:]) != [image_width, image_height]:
+            masks = masker(masks, prediction)
         # logger.info('Time mask: {}'.format(time.time() - t))
         # prediction = prediction.convert('xywh')
 
--- a/maskrcnn_benchmark/engine/inference.py
+++ b/maskrcnn_benchmark/engine/inference.py
@@ -87,10 +87,10 @@ def prepare_for_coco_segmentation(predictions, dataset):
         image_height = dataset.coco.imgs[original_id]["height"]
         prediction = prediction.resize((image_width, image_height))
         masks = prediction.get_field("mask")
-        
+
         # t = time.time()
         # Masker is necessary only if masks haven't been already resized.
-        if list(masks.shape[-2:]) != [image_width, image_height]:
+        if list(masks.shape[-2:]) != [image_height, image_width]:
             masks = masker(masks, prediction)
         # logger.info('Time mask: {}'.format(time.time() - t))
         # prediction = prediction.convert('xywh')
@@ -429,6 +429,6 @@ def inference(
     check_expected_results(results, expected_results, expected_results_sigma_tol)
     if output_folder:
         torch.save(results, os.path.join(output_folder, "coco_results.pth"))
-        
+
     return results, coco_results, predictions
 
--- a/maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
+++ b/maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
@@ -140,7 +140,7 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
     else:
         # for visualization and debugging, we also
         # allow it to return an unmodified mask
-        mask = mask * 255
+        mask = (mask * 255).to(torch.uint8)
 
     im_mask = torch.zeros((im_h, im_w), dtype=torch.uint8)
     x_0 = max(box[0], 0)
--- a/maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
+++ b/maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
@@ -120,7 +120,7 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
     padded_mask, scale = expand_masks(mask[None], padding=padding)
     mask = padded_mask[0, 0]
     box = expand_boxes(box[None], scale)[0]
-    box = box.type(dtype=torch.int32)
+    box = box.to(dtype=torch.int32)
 
     TO_REMOVE = 1
     w = box[2] - box[0] + TO_REMOVE
@@ -132,16 +132,12 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
     mask = mask.expand((1, 1, -1, -1))
 
     # Resize mask
-    mask = mask.type(torch.float32)
+    mask = mask.to(torch.float32)
     mask = F.interpolate(mask, size=(h, w), mode='bilinear')
     mask = mask[0][0]
 
     if thresh >= 0:
-        mask = (mask > thresh).type(torch.uint8)
-    else:
-        # for visualization and debugging, we also
-        # allow it to return an unmodified mask
-        mask = (mask * 255).type(torch.uint8)
+        mask = mask > thresh
 
     im_mask = torch.zeros((im_h, im_w), dtype=torch.uint8)
     x_0 = max(box[0], 0)
--- a/maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
+++ b/maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
@@ -44,11 +44,11 @@ class MaskPostProcessor(nn.Module):
         index = torch.arange(num_masks, device=labels.device)
         mask_prob = mask_prob[index, labels][:, None]
 
-        if self.masker:
-            mask_prob = self.masker(mask_prob, boxes)
-
         boxes_per_image = [len(box) for box in boxes]
         mask_prob = mask_prob.split(boxes_per_image, dim=0)
+
+        if self.masker:
+            mask_prob = self.masker(mask_prob, boxes)
 
         results = []
         for prob, box in zip(mask_prob, boxes):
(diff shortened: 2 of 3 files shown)


maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
```diff
@@ -119,25 +120,28 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
...
+
+    # Resize mask
+    mask = mask.type(torch.float32)
+    mask = F.interpolate(mask, size=(h, w), mode='bilinear')
+    mask = mask[0][0]
 
     if thresh >= 0:
-        mask = np.array(mask > thresh, dtype=np.uint8)
-        mask = torch.from_numpy(mask)
+        mask = (mask > thresh).type(torch.uint8)
```
fmassa: same comment about cast to `uint8` not being needed here.

maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
```diff
@@ -119,25 +120,28 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
...
     h = max(h, 1)
 
-    mask = Image.fromarray(mask.cpu().numpy())
-    mask = mask.resize((w, h), resample=Image.BILINEAR)
-    mask = np.array(mask, copy=False)
+    # Set shape to [batchxCxWxH]
+    mask = mask.expand((1, 1, -1, -1))
+
+    # Resize mask
+    mask = mask.type(torch.float32)
```
fmassa: Can we use instead `mask.to(torch.float32)` here? It's the new API for it, and `.type` is kind of deprecated.

maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
```diff
@@ -119,25 +120,28 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
     padded_mask, scale = expand_masks(mask[None], padding=padding)
     mask = padded_mask[0, 0]
     box = expand_boxes(box[None], scale)[0]
-    box = box.numpy().astype(np.int32)
+    box = box.type(dtype=torch.int32)
```
fmassa: can we use instead `box.to(torch.int32)`?

maskrcnn_benchmark/engine/inference.py
```diff
@@ -87,8 +87,11 @@ def prepare_for_coco_segmentation(predictions, dataset):
         image_height = dataset.coco.imgs[original_id]["height"]
         prediction = prediction.resize((image_width, image_height))
         masks = prediction.get_field("mask")
+        
         # t = time.time()
-        masks = masker(masks, prediction)
+        # Masker is necessary only if masks haven't been already resized.
+        if list(masks.shape[-2:]) != [image_width, image_height]:
```
fmassa: I think this is a bit fragile. Ideally, we would want the `Masker` to return a sequence of `Mask` objects, but this is not currently fully implemented, so we might want to let it for later.

Also, I think this should be `[image_height, image_width]`, as the shape of tensors is represented with `[N,C,H,W]` order generally.

hadim: Yes, ideally we would like Masker to be able to handle multi images batch.

hadim:
You're right about interpolation being done on float32. I'll try to run some benchmarks (probably not before next week).

But we could still keep the Numpy version and use the Torch one only if `POSTPROCESS_MASKS` is set to `True`.

hadim:
(By the way, it had to be said: your maskrcnn implementation is really great. The code logic and model structure is well designed and flexible enough. I have been looking for this kind of implementation for a long time since I don't have the time myself to implement it. Congratulation!)

maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
```diff
@@ -3,6 +3,7 @@
 import torch
 from PIL import Image
```
fmassa: I think this can now be removed, right?

maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
```diff
@@ -119,25 +120,24 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
...
+    mask = F.interpolate(mask, size=(h, w), mode='bilinear')
+    mask = mask[0][0]
 
     if thresh >= 0:
-        mask = np.array(mask > thresh, dtype=np.uint8)
-        mask = torch.from_numpy(mask)
-    else:
-        # for visualization and debugging, we also
-        # allow it to return an unmodified mask
-        mask = torch.from_numpy(mask * 255).to(torch.uint8)
```
fmassa: we still need this implementation for the demo, when we [pass `--show-mask-heatmaps`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/demo/webcam.py#L33), if not the mask will be a `float32` tensor with values from 0-1, while the rest of the code is expecting `uint8` in 0-255 (for visualization, or else we will have an all-black mask).

maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
```diff
@@ -119,25 +120,24 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
...
     TO_REMOVE = 1
     w = box[2] - box[0] + TO_REMOVE
     h = box[3] - box[1] + TO_REMOVE
     w = max(w, 1)
     h = max(h, 1)
 
-    mask = Image.fromarray(mask.cpu().numpy())
-    mask = mask.resize((w, h), resample=Image.BILINEAR)
-    mask = np.array(mask, copy=False)
+    # Set shape to [batchxCxWxH]
```
fmassa: nit: `[batchxCxHxW]`

fmassa:
And about keeping the PIL version, here is what I propose:

I have a CUDA and PyTorch-only implementation for the `paste_mask_in_image` which can run on either the CPU and the GPU.
I removed it from the current version of the code because it had a few boundary artifacts that dropped the mAP by a couple of points, and I didn't have the time to fix it.

I could share this PyTorch / CUDA implementation here, and we could look into fixing the boundary artifacts so that it gives the exact same results as before, and still allow to paste the masks on the GPU in parallel (which would be a win).

Thoughts?

maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
```diff
@@ -119,25 +119,28 @@ def paste_mask_in_image(mask, box, im_h, im_w, thresh=0.5, padding=1):
...
 
     if thresh >= 0:
-        mask = np.array(mask > thresh, dtype=np.uint8)
-        mask = torch.from_numpy(mask)
+        mask = mask > thresh
     else:
         # for visualization and debugging, we also
         # allow it to return an unmodified mask
-        mask = torch.from_numpy(mask * 255).to(torch.uint8)
+        mask = mask * 255
```
fmassa: This also needs to be casted to `uint8`, or else the line below will crash because we will be trying to copy a float tensor into a uint8 tensor.

hadim:
I've fixed issues from your comments.

Here is a notebook showing the Numpy version looks faster by 20-40%. I didn't do an extensive benchmark on a large image, so the reality could be different. Anyway, it's a good starting point for you if you want to test/benchmark this function (why not integrating it as a test in the repo?).

hadim:
Here is the notebook: https://gist.github.com/hadim/43e929498b8f84c6a19ccd62e4ff810a

hadim:
As for the CUDA version, let's integrate it in this PR. Better to do that properly rather than going back to this later.

As for the boundary artifacts, I am afraid that resizing the masks will always give a smaller mAP than without resizing no matter how good your resizing function is. Now the question is by how much the mAP is decreased... Only tests will tell us.

Before merging we also need to make `Masker` batch compatible don't you think? It's a pretty big limitation not being able to run inference on batch > 1.

fmassa:
No worries about training on COCO, we can do that here.
Note though that this function is used only during testing.

When I tried using the CUDA implementation from last time, it gave roughly 2 mAP lower than the current implementation. When I checked, it was mostly the boundaries that were a bit off by a few pixels.

Here is a gist with the CUDA (and equivalent CPU implementation): https://gist.github.com/fmassa/2b6763272ec867b6e7285684095b753d
The CUDA version is basically doing the same thing as the `compute_flow_field_cpu` function.

Unfortunately the CUDA version will probably not compile anymore due to a number of breaking changes in PyTorch internals, but those can be adapted without too much trouble so that it still compiles.

The idea is to create a warp mask which will be used to apply the `F.grid_sampler` on it.
The warp mask (or flow field) is -2 everywhere, except in the locations where the masks are present. In those places, we create an interpolation grid of the right size.

Let me know if you have questions. I'd recommend first trying to get the python implementation in `compute_flow_field_cpu` match what we currently have, and then worry about the GPU implementation.

hadim:
Could you check this notebook: https://gist.github.com/hadim/99181210e5788ae20c2a3bde72d268d1

In my test, the difference between the flow field CPU version and the previous one (`paste_mask_in_image`) is very small.

I don't know if it still too big for you or maybe the Pytorch `grid_sample()` function has been improved.

I almost didn't modify the source code (only some minor changes to update the aginst Pytorch 1.0 API).

hadim:
Also, the flow field version looks a bit faster than the previous one. I am waiting for your green light to integrate it into this PR.

fmassa:
Hum, interesting.

I'd need to run a full testing to see if it brings the mAP for masks down or not. I can do this next week.

We should also note that this `Masker` is applied in the probability maps, and not the thresholded images. There might be some more difference coming from there.

I'd prefer to have this change in a separate PR though. It will make merging this PR easier and quicker, if you are ok with it?

And it's going to be simpler for tracking the history. Could you maybe send another PR with this implementation?

hadim:
The last commit make Masker batch compatible. Backward compatibility is maintened.

hadim:
My solution raises an error during inference because masker returns a list instead of a tensor. I am not sure the returned masks could be merged into a single tensor since they don't have the same shape (number of objects. Or do they?

fmassa:
The line [just after the masker is applied](https://github.com/facebookresearch/maskrcnn-benchmark/blob/1890a84482a816b3f0b7453c2613c671932544c2/maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py#L51) converts the tensor into a list of tensors, one per each image.
So we would need to have some casing to split the `mask_prob` if the masker is not passed, or else use directly the result of Masker.
Something like

```python
        if self.masker:
            mask_prob = self.masker(mask_prob, boxes)
        else:
            mask_prob = mask_prob.split(boxes_per_image, dim=0)

        boxes_per_image = [len(box) for box in boxes]
```
might be enough.

Thoughts?

hadim:
I am confused because in `MaskPostProcessor`, the inputs of `forward()` are a list of BoxList, one element for each image and x for the masks.

To what corresponds to the first dimension of x?

hadim:
It looks like the first dimension is the number of images in the batch multiplied by the number of detected objects. Am I right?

hadim:
I think it's good to be reviewed.

maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py
```diff
@@ -44,11 +60,13 @@ def forward(self, x, boxes):
         index = torch.arange(num_masks, device=labels.device)
         mask_prob = mask_prob[index, labels][:, None]
 
+        boxes_per_image = [len(box) for box in boxes]
+        
         if self.masker:
+            mask_prob = split_list(mask_prob, boxes_per_image)
```
fmassa: I'm confused: isn't `mask_prob` a single tensor at this stage? If yes, why can't we use `.split` here?

fmassa:
Hi @hadim 

I'll try finding some time late this week or early next week to pull your changes and test some things locally so that I can merge this.

Sorry for the delay!

fmassa:
Thanks a lot @hadim , I've merged the changes in #180 (I couldn't push the changes to this PR)

Issue #128, opened 2018-11-07
Add configs for  PASCAL VOC datasets


lufficc:
## 🚀 Feature
It will be very convenient to add configs for  PASCAL VOC datasets, as PASCAL VOC datasets are smaller and easily trained.

fmassa:
I agree.

We would also need to have a more generic way of computing the mAP that doesn't require `cocoAPI`, or have the VOC datasets converted into COCO format.

PR #131, opened 2018-11-08
add support for pascal voc dataset and evaluate

lufficc:
See  [#128](https://github.com/facebookresearch/maskrcnn-benchmark/issues/128#issue-378269486).


maskrcnn_benchmark/engine/inference.py
```diff
@@ -6,16 +6,18 @@
...
 from collections import OrderedDict
 
-import torch
+import numpy as np
 
+from ..config import cfg
+import torch
 from tqdm import tqdm
-
+from ..data.datasets.voc import VOC_BBOX_LABEL_NAMES
```
fmassa: Ideally we would keep the inference engine agnostic to particular dataset implementations. But this will require some refactorings.

maskrcnn_benchmark/engine/inference.py
```diff
@@ -6,16 +6,18 @@
 import os
 from collections import OrderedDict
 
-import torch
+import numpy as np
 
+from ..config import cfg
```
fmassa: I'd rather not import the config in this file, but keep it agnostic to it. You can probably pass the name of the dataset as an argument to the `inference` function

maskrcnn_benchmark/data/datasets/voc_eval.py
```diff
@@ -0,0 +1,324 @@
...
+    # bottom right
+    br = xp.minimum(bbox_a[:, None, 2:], bbox_b[:, 2:])
+
+    area_i = xp.prod(br - tl, axis=2) * (tl < br).all(axis=2)
+    area_a = xp.prod(bbox_a[:, 2:] - bbox_a[:, :2], axis=1)
+    area_b = xp.prod(bbox_b[:, 2:] - bbox_b[:, :2], axis=1)
+    return area_i / (area_a[:, None] + area_b - area_i)
+
+
+def eval_detection_voc(
```
fmassa: If those functions come from an already-existing repo, it would be good to reference them here.

maskrcnn_benchmark/engine/inference.py
```diff
@@ -386,46 +423,54 @@ def inference(
...
-    logger.info(results)
-    check_expected_results(results, expected_results, expected_results_sigma_tol)
-    if output_folder:
-        torch.save(results, os.path.join(output_folder, "coco_results.pth"))
-        
-    return results, coco_results, predictions
-
+            torch.save(results, os.path.join(output_folder, "coco_results.pth"))
+        return results, coco_results
+    if 'coco' in cfg.DATASETS.TEST[0]:
```
fmassa: This doesn't work if we pass multiple datasets during evaluation. I think it might be better to pass a dataset name as an argument to the `inference` function

maskrcnn_benchmark/data/datasets/voc_eval.py
```diff
@@ -0,0 +1,324 @@
+from __future__ import division
+
+from collections import defaultdict
+import itertools
+import numpy as np
+import numpy as xp
```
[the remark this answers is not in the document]
lufficc: It is aduplicated import, I've optimized it.

maskrcnn_benchmark/data/datasets/voc.py
```diff
@@ -0,0 +1,118 @@
...
+
+    def __getitem__(self, index):
+        img_id = self.ids[index]
+        img = Image.open(os.path.join(self.imgs_dir, '%s.jpg' % img_id)).convert('RGB')
+        ann = self.anns[img_id]
+        boxes = ann['boxes']
+        labels = ann['labels']
+
+        boxes = torch.as_tensor(boxes).reshape(-1, 4)  # guard against no boxes
+        target = BoxList(boxes, img.size, mode="xyxy").convert("xyxy")
```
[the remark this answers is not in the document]
lufficc: I've optimized it.

maskrcnn_benchmark/data/datasets/voc.py
```diff
@@ -0,0 +1,118 @@
+import os
+
+import torch
+import torch.utils.data as data
+from PIL import Image
+from lxml import etree
+
+from maskrcnn_benchmark.structures.bounding_box import BoxList
+
+VOC_BBOX_LABEL_NAMES = ('aeroplane',
```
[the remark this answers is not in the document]
lufficc: Great advise. I've optimized it.

maskrcnn_benchmark/data/datasets/voc.py
```diff
@@ -0,0 +1,118 @@
...
+            self.ids = sorted([l.strip() for l in fid.readlines()])
+
+        self.json_category_id_to_contiguous_id = {
+            v: i + 1 for i, v in enumerate(range(20))
+        }
+        self.contiguous_category_id_to_json_id = {
+            v: k for k, v in self.json_category_id_to_contiguous_id.items()
+        }
+        self.anns = {}
+        for img_id in self.ids:
```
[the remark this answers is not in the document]
lufficc: Infact it costs very little time(<1s).To keep consistent, I add a print to tell the cost time like coco api does.

maskrcnn_benchmark/data/datasets/voc_eval.py
```diff
@@ -0,0 +1,324 @@
...
+import six
+
+
+def bbox_iou(bbox_a, bbox_b):
```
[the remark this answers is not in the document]
lufficc: To keep consistent, I removed bbox_iou and used structures/boxlist_ops.py's boxlist_iou.

maskrcnn_benchmark/engine/inference.py
lufficc: I removed it.

lufficc:
Sorry, I cannot make a new pr, this pr has already contained the new changes. My mistakes.

maskrcnn_benchmark/engine/inference.py
lufficc: Removed.

maskrcnn_benchmark/engine/inference.py
```diff
@@ -366,7 +363,9 @@ def inference(
     )
     logger = logging.getLogger("maskrcnn_benchmark.inference")
     dataset = data_loader.dataset
-    logger.info("Start evaluation on {} images".format(len(dataset)))
+    assert hasattr(dataset, 'name'), 'Dataset must has a name to perform evaluating.'
+    dataset_name = dataset.name
```
fmassa: I was actually thinking about passing a `dataset_name` as an argument to `inference`, instead of adding a field in the dataset itself. What do you think?

maskrcnn_benchmark/data/datasets/voc.py
```diff
@@ -0,0 +1,117 @@
...
+            with open(os.path.join(self.anns_dir, '%s.xml' % img_id)) as fid:
+                xml_str = fid.read()
+            xml = etree.fromstring(xml_str)
+            data = self._recursive_parse_xml_to_dict(xml)['annotation']
+            self.anns[img_id] = {
+                'boxes': [[int(obj['bndbox']['xmin']),
+                           int(obj['bndbox']['ymin']),
+                           int(obj['bndbox']['xmax']),
+                           int(obj['bndbox']['ymax'])] for obj in data['object'] if int(obj['difficult']) == 0],
```
fmassa: I think this is not the best thing to do unconditionally.

Indeed, we might want to remove the difficult instances during training, but during testing there should be a way of knowing which one of the instances are difficult, so that they are not taken into account during evaluation.

If I understand it correctly, the way it is currently done will be penalizing the difficult instances, as if we detect a difficult instance with our model, it will be counted as a false match.

I think we should have a flag that lets us decide if we want to return difficult instances or not.

maskrcnn_benchmark/data/datasets/coco.py
```diff
@@ -8,10 +8,10 @@
 
 class COCODataset(torchvision.datasets.coco.CocoDetection):
     def __init__(
-        self, ann_file, root, remove_images_without_annotations, transforms=None
+        self, ann_file, root, name, remove_images_without_annotations, transforms=None
     ):
         super(COCODataset, self).__init__(root, ann_file)
-
+        self.name = name
```
fmassa: Let's maybe remove the `name` attribute in favour of the approach I mentioned just before?

What are your thoughts?

maskrcnn_benchmark/engine/inference.py
```diff
@@ -386,6 +385,35 @@ def inference(
...
+                                    output_folder=output_folder,
+                                    logger=logger)
+        coco_results = None
+    else:
+        raise NotImplementedError('Unsupported dataset type %s.' % dataset_name)
+
+    return results, coco_results, predictions
+
+
+def do_coco_evaluation(predictions,
```
fmassa: What do you think about moving the coco-specific and pascal-specific evaluation functions into their own file, and have the inference file here be very agnostic to the dataset? It is almost there, right?

maskrcnn_benchmark/data/datasets/voc_eval.py
```diff
@@ -0,0 +1,186 @@
+# A modification version from chainercv repository.
+# (See https://github.com/chainer/chainercv/blob/master/chainercv/evaluations/eval_detection_voc.py)
+from __future__ import division
```
fmassa: I wonder if this file should live in `datasets`, or in a separate folder dedicated to inference. Thoughts?

lufficc: In fact, `inference.py` looks a little chaotic.
I think  `inference.py` should only do the predication, then pass results and call different evaluatation methods according to the dataset_name. Like you mentioned before.

The location of 'different evaluatation methods', I think, can be:
```
datasets
  -evaluatation
    -coco
      ...
      ...
    -voc
     ...
     ...
   coco.py
   voc.py
```

What do you think?

maskrcnn_benchmark/data/datasets/coco.py
lufficc: I got it. This would be a better choice.

maskrcnn_benchmark/data/datasets/voc.py
lufficc: I got it.

maskrcnn_benchmark/data/datasets/voc_eval.py
fmassa: I agree with you, let's keep this structure that you proposed.

lufficc: I've make a new commit. Thanks for your review again.

maskrcnn_benchmark/data/datasets/evaluation/voc/voc_eval.py
```diff
@@ -0,0 +1,220 @@
...
+        gt_boxlists(list[BoxList]): ground truth boxlist, has labels field.
+        iou_thresh: iou thresh
+        use_07_metric: boolean
+    Returns:
+        dict represents the results
+    """
+    assert len(gt_boxlists) == len(pred_boxlists), 'Length of gt and pred lists need to be same.'
+    prec, rec = calc_detection_voc_prec_rec(pred_boxlists=pred_boxlists,
+                                            gt_boxlists=gt_boxlists,
+                                            iou_thresh=iou_thresh)
```
fmassa: You still need to pass `gt_difficults` in here if you want it to be evaluated properly

lufficc:
If I load the xml at every iteration, there may be a duplicated loading(when call `get_img_info`). So I think a better choice is load all annotations(include difficult objects) into memory, then filter them dynamically(according to `use_difficult` flag).

fmassa:
This is looking much better. I'd rename `difficulties` to `difficult` though, but that's minor.

I'll try using the code and training a model in the weekend / next week and once I've verified that everything is working as expected I'll merge the PR or make further comments.

Thanks!

lufficc:
OK. Thanks!

henrywang1:
Hi,
Thanks for the wonderful commit and review.
I was also trying different datasets on maskrcnn-benchmark, and I have some comments.

1. I think we could load the notation from json files directly like Detectron, it has [VOC Annotations in COCO Format].(https://github.com/facebookresearch/Detectron/tree/8170b25b425967f8f1c7d715bea3c5b8d9536cd8/detectron/datasets/data)
2. And we could also evaluate the segmentation result on voc dataset. (The information is also in the json files.)
What do you think? Maybe I could create another issue?

Thanks!

fmassa:
Hi @henrywang1 ,
Yesterday I pulled out this repo and I started doing some minor changes, I should get this ready by tomorrow.

About your comments:
1 - yes, it is possible to use COCO annotations for Pascal VOC dataset. But I think it is good to have built-in support for some evaluation code which doesn't depend on COCO, and this PR adds that
2 - indeed, it could be interesting to add. Do you know if we generally evaluate instance segmentation on Pascal as well? I know that semantic segmentation is commonly evaluated, but I'm not sure about instance segmentation.

Thanks

henrywang1:
Hi @fmassa 
As we know, COCO is a more commonly used dataset for instance segmentation.
But there are still some recent papers evaluating their instance segmentation models on PASCAL.
I list some of the papers for your reference.
- https://arxiv.org/abs/1611.07709
- https://arxiv.org/abs/1703.10277
- https://arxiv.org/abs/1804.00880

In my opinion, the feature could benefit people who want to try Mask-RCNN, but they don't have much computing power.

Thanks

fmassa:
Sounds good, it would be great to add support for instance masks as well on Pascal at a later PR.

But that might mean also implementing the evaluation code, or just directly using the data in COCO format

lufficc:
I've checked "Allow edits from maintainers", do you have permission now?

fmassa:
@lufficc I couldn't make it work as I was expecting, so I opened a new PR in #207 .
It contains all your diffs, plus a few more from me.

I'm very sorry, I need to learn how to do this properly... :-/

fmassa:
I was following instructions from https://help.github.com/articles/committing-changes-to-a-pull-request-branch-created-from-a-fork/ , but I definitely did something wrong...

lufficc:
It's OK, I'm not very familiar with that either 😆 . What should I do next?

fmassa:
Maybe you could have a look at #207 (or more precisely the diffs that I added) and see if you are ok with that?

fmassa:
Maybe the problem happened because your diff was based on the master branch, and not on a separate branch? But I'm not 100% sure

lufficc:
I'm OK with your new commits. 
Or maybe I didn't merge the pr with the master's HEAD?

fmassa:
I don't actually know. I did almost exactly the same thing with https://github.com/facebookresearch/maskrcnn-benchmark/pull/102#issuecomment-441263117 but it worked out fine in the end.

zimenglan-sysu-512:
hi @fmassa 
will u add the results of pascal voc into `MODEL_ZOO.md`?

fmassa:
Maybe, but I'd need to train a few more models and compare against what is currently available, so this will probably not happen in the near future

narendoraiswamy:
Hello @fmassa Correct me if  I am wrong but support for pascal voc dataset for instance segmentation has not been completed yet?  Using the `voc.py` dataloader only supports the detection purpose and not the instance segmentation purpose.

PR #140, opened 2018-11-11
Return proposals during training to allow evaluation.

hadim:
It allows performing evaluation during training.

### DIFF
--- a/maskrcnn_benchmark/engine/trainer.py
+++ b/maskrcnn_benchmark/engine/trainer.py
@@ -62,7 +62,7 @@ def do_train(
         images = images.to(device)
         targets = [target.to(device) for target in targets]
 
-        loss_dict = model(images, targets)
+        loss_dict, _ = model(images, targets)
 
         losses = sum(loss for loss in loss_dict.values())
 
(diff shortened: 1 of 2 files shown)


fmassa:
We can use `not self.training` for evaluation.

I think that having a separate process (potentially in a separate machine) is maybe more desirable. This was something that has been asked already, and the original Detectron has such functionality, so I think it might be a good thing to follow.

hadim:
Ok, no problem. Everyone can have a custom training loop anyway.

But in that case, are you suggest to have a new `model` object used only for evaluation?

fmassa:
If you want to use some model for evaluation, just put it in `model.eval()` mode before starting the evaluation, and move it back to `model.train()` mode just before the training.

Another possibility would be to enable to have some hooks that can be attached to the training job, and that could enable the evaluation to be performed at every x iterations.
But care should be taken when using multiple GPUs with distributed data parallel (like unwrapping the `DistributedDataParallel` model for inference, and wrapping it back for training).

hadim:
Ok I see. I'll use `model.train()` and `model.eval()` then. Thank you. I'll close here.

fmassa:
Thanks for the PRs and the contributions!

PR #145, opened 2018-11-13
Fix iteration count

jimmyyhwu:
This commit changes the iteration count to reflect the true number of iterations a model has been trained for. Previously, a model trained for n + 1 iterations would be saved as model_n.pth.

This also fixes the issue where resuming from a checkpoint would result in the model training for one iteration and then immediately saving to the same checkpoint file it resumed from.


jimmyyhwu:
Yes, that should work. I modified the commit to reflect the changes.

PR #150, opened 2018-11-13
support RLE and binary mask

wangg12:


maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -1,51 +1,97 @@
...
+        TO_REMOVE = 1
+        w, h = box[2] - box[0] + TO_REMOVE, box[3] - box[1] + TO_REMOVE
+        cropped_mask = self.mask[box[1]: box[3]+1, box[0]: box[2]+1]
+        return Mask(cropped_mask, size=(w, h), mode=self.mode)
 
+    # torch.nn.functional.interpolate has a arg as dim, only tensor have dim, so turn array to tensor
     def resize(self, size, *args, **kwargs):
-        pass
+        width, height = size
+        scaled_mask = torch.squeeze(interpolate(torch.from_numpy(np.array(self.mask)[None, None, :, :]).float(),
```
fmassa: Why do you perform this numpy conversion? torch tensors also support the `newaxis` style, and this can also be done via
```python
scaled_mask = interpolate(self.mask[None, None, :, :])[0, 0]
```

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -90,7 +136,8 @@ def transpose(self, method):
         return Polygons(flipped_polygons, size=self.size, mode=self.mode)
 
     def crop(self, box):
-        w, h = box[2] - box[0], box[3] - box[1]
+        TO_REMOVE = 1
+        w, h = box[2] - box[0] + TO_REMOVE, box[3] - box[1] + TO_REMOVE
```
fmassa: Indeed, this used to look inconsistent, but I think [it mimics the behavior of Detectron in this part of the code](https://github.com/facebookresearch/Detectron/blob/master/detectron/utils/segms.py#L110), so I'd maybe rather keep a comment in here instead of modifying the code, as it will change the accuracies (for the better or for the worse)?

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -130,7 +177,8 @@ def convert(self, mode):
             )
             rle = mask_utils.merge(rles)
             mask = mask_utils.decode(rle)
-            mask = torch.from_numpy(mask)
+            mask = np.sum(mask, axis=2)
```
fmassa: Doesn't this have problems with the rest of the code? Like during training of Mask R-CNN, we call the convert function and that expects a tensor of a particular number of dimensions. I'm afraid this will break the code there.

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -169,46 +226,46 @@ def transpose(self, method):
...
         flipped = []
-        for polygon in self.polygons:
-            flipped.append(polygon.transpose(method))
+        for mask in self.masks:
+            flipped.append(mask.transpose(method))
         return SegmentationMask(flipped, size=self.size, mode=self.mode)
 
     def crop(self, box):
-        w, h = box[2] - box[0], box[3] - box[1]
+        w, h = box[2] - box[0] + 1, box[3] - box[1] + 1
```
fmassa: Same comment as before about the +1, can we keep it as it was and add a comment instead?

wangg12: I think +1 is the correct equation, because you didn't use the `w, h` in the class `Polygons`, the performance is actually unrelated to that.

maskrcnn_benchmark/structures/segmentation_mask.py
fmassa: Yes, this is supposed to be the conversion from one format to the other, but there is an inconsistency in Detectron I believe, and I tried to keep this inconsistency here for the sake of reproducibility

maskrcnn_benchmark/structures/segmentation_mask.py
fmassa: I think I'm missing something here, but how this won't be a breaking change compared to the previous code, as this is used in [here](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/modeling/roi_heads/mask_head/loss.py#L40) to compute the loss?

maskrcnn_benchmark/structures/segmentation_mask.py
fmassa: I think a good exercise here is to try to make in such a way that [this part of the code in Detectron](https://github.com/facebookresearch/Detectron/blob/master/detectron/utils/segms.py#L110-L121) is equivalent to [this part in maskrcnn-benchmark](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/modeling/roi_heads/mask_head/loss.py#L38-L39).
I think that for them to be equivalent, we should not have the +1, but I might be wrong here.

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: Could you please point me out where the inconsistency happens in Detectron?

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: I mimics the code in Detectron ([here](https://github.com/facebookresearch/Detectron/blob/master/detectron/utils/segms.py#L84)). I didn't test it using polygons format. If this is wrong, I will remove it. I'm a little confused about why `sum` is not needed here.

fmassa: Well, this was working before, so this will probably crash now :-)
I don't remember all the details now and I'd need to check the difference. But maybe you can have a look at the code and run it to see what it gives?

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: OK. I think for the resize operation in polygons, there should be no `+1`. But the crop operation should have `+1`.

maskrcnn_benchmark/structures/segmentation_mask.py
fmassa: In detectron, they use the +1 everywhere, except in the masks in the point I mentioned to you. Have a look at my other comment for a bit more info

fmassa:
Also, I think which would be awesome to have would be to add some tests in `maskrcnn-benchmark/tests`, which we are currently lacking and would make reviewing the changes much easier!

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: I just test the code on my RLE format dataset. It will take me some time to test if on coco.

wangg12:
@fmassa I think it should be consistent with Detectron now.

fmassa:
I'll have another closer look today, but would you mind adding some tests in `tests/`? It will make things much easier!

wangg12:
@fmassa I'm not familiar with how to write the tests.

fmassa:
Ok, no worries

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -1,51 +1,93 @@
...
+    def __init__(self, segm, size, mode):
+        width, height = size
+        if isinstance(segm, Mask):
+            mask = segm.mask
+        else:
+            if type(segm) == list:
+                # polygons
+                rle = mask_utils.frPyObjects(segm, height, width)
+                mask = np.array(mask_utils.decode(rle), dtype=np.float32)
+                mask = np.sum(mask, axis=2)
```
fmassa: Is this sum here needed?

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -1,51 +1,93 @@
...
     """
-
-    def __init__(self, masks, size, mode):
-        self.masks = masks
+    def __init__(self, segm, size, mode):
+        width, height = size
+        if isinstance(segm, Mask):
+            mask = segm.mask
+        else:
+            if type(segm) == list:
```
fmassa: I'm not sure we want to support all those different kinds of data in the `Mask` class, specially that we can't convert back and forth between them.
What about letting everything polygon-based (including RLE?) be handled by the Polygon class, and call .convert here if it is a Polygon instance?

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -1,51 +1,93 @@
...
+        flipped_mask = self.mask.index_select(dim, flip_idx)
+        return Mask(flipped_mask, self.size, self.mode)
 
     def crop(self, box):
+        box = [int(b) for b in box]
         w, h = box[2] - box[0], box[3] - box[1]
-
-        cropped_masks = self.masks[:, box[1] : box[3], box[0] : box[2]]
-        return Mask(cropped_masks, size=(w, h), mode=self.mode)
+        cropped_mask = self.mask[box[1]: box[3]+1, box[0]: box[2]+1]
```
fmassa: The indexing here depends on the data: we should have a convention to add the +1 or not. I'm undecided about it for the moment. Ideally, training with polygons or masks should give the same result, and I'm unsure this is the case now

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: I think it is need according to the code in Detectron and my experience. A binary mask may contain multiple polygons and to convert polygons into binary mask, the third dimension should be merged by summation.

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: Cropping on polygons does not need the `w,h`, but I think this is necessary here.

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: Current `convert` method in `Polygons` class does not always return a single channel mask. So I'm not sure whether it could be used here. Another option is to write a function outside for converting the format and handle the types.

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: Sounds good to me. I will verify this.

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: `sum` here is equivalent to `merge` in `Pylygons.convert`.

wangg12:
@fmassa I've added some tests about segmentation_mask. However, the `transpose` and `resize` can not pass the test. Some help is needed.

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: @fmassa You're right about this. I've verified through tests.

tests/common_utils.py
```diff
@@ -0,0 +1,839 @@
+r"""Importing this file must **not** initialize CUDA context. test_distributed
```
fmassa: I don't think we need all this file and the next one (which comes from PyTorch I suspect?) here.
Can't we just do some basic functionality like in https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/tests/test_data_samplers.py ?

tests/test_segmentation_mask.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def test_transpose(self):
+        FLIP_LEFT_RIGHT = 0
+        FLIP_TOP_BOTTOM = 1
+        methods = (FLIP_LEFT_RIGHT, FLIP_TOP_BOTTOM)
+        for method in methods:
+            mask_from_poly_flip = self.polygon.transpose(method).convert('mask')
+            mask_flip = self.mask.transpose(method).convert('mask')
+            print(method, torch.abs(mask_flip.float() - mask_from_poly_flip.float()).sum())
+            self.assertEqual(mask_flip, mask_from_poly_flip)
```
fmassa: Because we won't have support for torch tensors once we remove `common_utils`, I propose we use something like
```python
 self.assertTrue(torch.equal(tensor1, tensor2))

```

wangg12:
For `resize` I can understand there maybe some numerical difference because of the interpolation, but I don't know how to make them equivalent. For `transpose` the difference is much stranger, I don't know where the difference is.

fmassa:
For the `resize`, I'd check if the difference appears in the boundaries. Also, the values should be 0-1.
For the `transpose`, I'd see if the difference disappears if you [set this `TO_REMOVE = 0`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/7f22baab128958e07977fd6e1dcb92a097722477/maskrcnn_benchmark/structures/segmentation_mask.py#L127). That could explain the difference.

Also, visualizing both results in the image space is going to be very helpful.

wangg12:
@fmassa For `transpose`, I changed `TO_REMOVE=0` in class `Polygons`. The difference is smaller but still exists. For `resize`, I've visualized the result, the difference almost comes from the boundaries.

fmassa:
Ok, this is good progress, thanks!

I'd need to look more closely to see where the difference might come from for the `transpose`, there might be a few pixels off and this can be observed also by viewing the images.

For `resize`, I'd need to check a bit more carefully to see if I spot anything, but I might not have the time today nor tomorrow.

JoyHuYY1412:
> @fmassa For `transpose`, I changed `TO_REMOVE=0` in class `Polygons`. The difference is smaller but still exists. For `resize`, I've visualized the result, the difference almost comes from the boundaries.

If I use the `segmentation_mask.py` you wrote, are there other changes I should apply on other files?
It looks like the functions and ports fit the original code well. 
What's more, are there other things I should take attention? Maybe I can try the transforms without transposing first?  >< Thank you!

JoyHuYY1412:
> @JoyHuYY1412 you'd need to check the interpolation (to use bilinear instead of nearest), apart from that, the rest should be unchanged (or almost).
> Thank you~

IssamLaradji:
I like this!

IssamLaradji:
I faced a problem with `cropped_mask = self.mask[box[1]: box[3], box[0]: box[2]]` in the part of the code below where I had `box[0]` and `box[2]` equal, resulting in `width=0`, causing an error.
```
def crop(self, box):
        box = [int(b) for b in box]
        w, h = box[2] - box[0], box[3] - box[1]
        w = max(w, 1)
        h = max(h, 1)
        cropped_mask = self.mask[box[1]: box[3], box[0]: box[2]]
        return Mask(cropped_mask, size=(w, h), mode=self.mode)
```

wangg12:
@IssamLaradji I think it should be `w, h = box[2] - box[0] + 1, box[3] - box[1] + 1`.

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -148,17 +189,26 @@ class SegmentationMask(object):
...
+                (2) rles: COCO's run length encoding format, uncompressed or compressed
+                (3) binary masks
+            size: (width, height)
+            mode: 'polygon', 'mask'. if mode is 'mask', convert mask of any format to binary mask
         """
-        assert isinstance(polygons, list)
-
-        self.polygons = [Polygons(p, size, mode) for p in polygons]
+        assert isinstance(segms, list)
+        if type(segms[0]) != list:
```
ZENGXH:
```suggestion
        if type(segms[0]) != list and type(segms[0]) != Polygons:
```
There is an error when I run training on voc/coco with this `SegmentationMask`: 
after calling `target = target.clip_to_image(remove_empty=True)` in `coco.py`, segms[0] will be Polygon, it shouldn't be assigned as 'Mask' in that case.

fmassa:
@botcs if you could write unit tests for this PR, it would be awesome!

botcs:
using the available [`test_segmentation_mask`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/b8c5bcafa02e4a7489b5a79bf4609933f98c9599/tests/test_segmentation_mask.py), I have tried to visualize the Polygon and the Mask tests, and here is a notebook of my observations:
https://gist.github.com/botcs/95176d877dcd26e48e46cceecdac5763

maskrcnn_benchmark/structures/segmentation_mask.py
JoyHuYY1412: yes thanks

fmassa:
@botcs that's awesome, thanks for the notebook!
So indeed we have the boundary effects as before.

One last thing I'd like to do to verify if this boundary effects actually matter is to run a training with https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/configs/quick_schedules/e2e_mask_rcnn_R_50_FPN_quick.yaml , which is a fast training and should take ~10 min to run, using the `Mask` class instead of Polygons.

For that, I'd (locally, this is not to be commited) modify https://github.com/facebookresearch/maskrcnn-benchmark/blob/f8b0118392aae793771cfc6cd869ad0a601d5c92/maskrcnn_benchmark/data/datasets/coco.py#L82-L84
so that it uses `Mask` instead.

I'd expect to have as results something like
```
EXPECTED_RESULTS: [['coco_2014_minival', 'box', 'AP', [0.082300, 0.001682]], ['coco_2014_minival', 'mask', 'AP', [0.075039, 0.001872]]]
```
, so box AP should be around 8.2, and mask AP around 7.5

Could you do that?

Thanks!

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -1,51 +1,92 @@
...
+        s = self.__class__.__name__ + "("
+        # s += "num_mask={}, ".format(len(self.mask))
+        s += "image_width={}, ".format(self.size[0])
+        s += "image_height={}, ".format(self.size[1])
+        s += "mode={})".format(self.mode)
+        return s
+
 
 
 class Polygons(object):
```
botcs: I think the `mode` field for the Polygon is completely irrelevant.
It is never used, but causes:
-  additional argument passing when constructing
- a headache when try to find out what Polygon really is

**Question:**
Wouldn't it be more consistent if the `convert` method of a Polygon would be renamed to `convert_to_mask` and would return a Mask instance?

fmassa: Hi,

I agree with your points.

The reason why this is currently the case was that I wanted to keep the same interface between `Polygons` and `Box` (which is not implemented, but is the single-box equivalent of `BoxList`).
And my original idea was that we would be able to specify what was the underlying type of the data via the mode: is it a polygon, or a mask?

I'm not sure about changing the `convert` name of the method though.

In general, I think both `box_list` and `segmentation_mask` could benefit from some better design / cleanup, but I'm not sure what that would be

botcs: > And my original idea was that we would be able to specify what was the underlying type of the data via the mode: is it a polygon, or a mask?

I think I cannot follow this part:
> To specify what was the underlying type of the data via the mode

As in the current implementation, a `Polygon` instance: 
1.  can be initialized either with a `list` of polygons
1.  can be initialized either with a `Polygon` instance (which is referenced now, but **should** be hard-copied IMO)
1.  **cannot** be initialized with a `Mask`, which feature could be added if necessary (I am doing this to convert GTA binary masks to COCO Polygon format, but only because the binary masks are not supported).

So the underlying data would be specified: Polygon.

On the other hand, about the `convert` function:
>I'm not sure about changing the `convert` name of the method though.
1. The `convert` function takes an argument for the target mode, but actually it accepts just a single answer, which is odd.
2. If I assume that a `Polygon` can be only `convert`-ed to a `Mask` than `convert` name is OK, but relies on the assumption that the data can be represented either in `Polygon`s or `Mask`s and nothing else, which is not necessary a trivial assumption, so changing the name to `convert_to_mask` would be clear from the very first encounter.

> keep the same interface
1. We should in this case add the `convert` or `convert_to_polygon` method to the `Mask` class as well

botcs:
> @botcs that's awesome, thanks for the notebook!
> So indeed we have the boundary effects as before.
> 
> One last thing I'd like to do to verify if this boundary effects actually matter is to run a training with https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/configs/quick_schedules/e2e_mask_rcnn_R_50_FPN_quick.yaml , which is a fast training and should take ~10 min to run, using the `Mask` class instead of Polygons.
> 
> For that, I'd (locally, this is not to be commited) modify
> 
> [maskrcnn-benchmark/maskrcnn_benchmark/data/datasets/coco.py](https://github.com/facebookresearch/maskrcnn-benchmark/blob/f8b0118392aae793771cfc6cd869ad0a601d5c92/maskrcnn_benchmark/data/datasets/coco.py#L82-L84)
> 
> Lines 82 to 84 in [f8b0118](/facebookresearch/maskrcnn-benchmark/commit/f8b0118392aae793771cfc6cd869ad0a601d5c92)
>  masks = [obj["segmentation"] for obj in anno] 
>  masks = SegmentationMask(masks, img.size) 
>  target.add_field("masks", masks) 
> 
> so that it uses `Mask` instead.
> 
> I'd expect to have as results something like
> 
> ```
> EXPECTED_RESULTS: [['coco_2014_minival', 'box', 'AP', [0.082300, 0.001682]], ['coco_2014_minival', 'mask', 'AP', [0.075039, 0.001872]]]
> ```
> 
> , so box AP should be around 8.2, and mask AP around 7.5
> 
> Could you do that?
> 
> Thanks!

I have trained the model, which went fine, but the evaluation has failed with the following error:
```
  File "/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/maskrcnn_benchmark-0.1-py3.7-linux-x86_64.egg/maskrcnn_benchmark/structures/segmentation_mask.py", line 206, in __init__
    if not isinstance(segms[0], (list, Polygons)):
IndexError: list index out of range
```

the whole output can be foun at this [gist](https://gist.github.com/botcs/5d60ac6cad83e9776afeeb0113cf4a2f)

botcs:
I have suppressed the error by using single empty instance with an all-zero mask when the provided `segms` is an empty list. And the results are the following:

```
2019-02-19 15:14:13,030 maskrcnn_benchmark.inference INFO: Start evaluation on coco_2014_minival dataset(5000 images).
  0%|▍                                                                                                                                                                         | 6/2500 [00:00<08:55,  4.66it/s]/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/torch/nn/functional.py:2423: UserWarning: Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0. Please specify align_corners=True if the old behavior is desired. See the documentation of nn.Upsample for details.
  "See the documentation of nn.Upsample for details.".format(mode))
  1%|█▍                                                                                                                                                                       | 21/2500 [00:01<03:42, 11.15it/s]/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/torch/nn/functional.py:2423: UserWarning: Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0. Please specify align_corners=True if the old behavior is desired. See the documentation of nn.Upsample for details.
  "See the documentation of nn.Upsample for details.".format(mode))
  1%|█▌                                                                                                                                                                       | 23/2500 [00:02<03:32, 11.64it/s]/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/torch/nn/functional.py:2423: UserWarning: Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0. Please specify align_corners=True if the old behavior is desired. See the documentation of nn.Upsample for details.
  "See the documentation of nn.Upsample for details.".format(mode))
  2%|███▏                                                                                                                                                                     | 47/2500 [00:03<03:07, 13.06it/s]/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/torch/nn/functional.py:2423: UserWarning: Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0. Please specify align_corners=True if the old behavior is desired. See the documentation of nn.Upsample for details.
  "See the documentation of nn.Upsample for details.".format(mode))
  4%|███████▌                                                                                                                                                                | 112/2500 [00:08<02:52, 13.81it/s]/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/torch/nn/functional.py:2423: UserWarning: Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0. Please specify align_corners=True if the old behavior is desired. See the documentation of nn.Upsample for details.
  "See the documentation of nn.Upsample for details.".format(mode))
 13%|█████████████████████▍                                                                                                                                                  | 319/2500 [00:24<02:44, 13.25it/s]/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/torch/nn/functional.py:2423: UserWarning: Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0. Please specify align_corners=True if the old behavior is desired. See the documentation of nn.Upsample for details.
  "See the documentation of nn.Upsample for details.".format(mode))
 33%|██████████████████████████████████████████████████████▊                                                                                                                 | 815/2500 [01:02<02:10, 12.96it/s]/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/torch/nn/functional.py:2423: UserWarning: Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0. Please specify align_corners=True if the old behavior is desired. See the documentation of nn.Upsample for details.
  "See the documentation of nn.Upsample for details.".format(mode))
 48%|███████████████████████████████████████████████████████████████████████████████▌                                                                                       | 1191/2500 [01:33<01:47, 12.17it/s]/home/csbotos/anaconda3/envs/debugmask/lib/python3.7/site-packages/torch/nn/functional.py:2423: UserWarning: Default upsampling behavior when mode=bilinear is changed to align_corners=False since 0.4.0. Please specify align_corners=True if the old behavior is desired. See the documentation of nn.Upsample for details.
  "See the documentation of nn.Upsample for details.".format(mode))
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2500/2500 [03:13<00:00, 13.15it/s]
100%|███████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 2500/2500 [03:20<00:00, 12.44it/s]
2019-02-19 15:17:34,026 maskrcnn_benchmark.inference INFO: Total inference time: 0:03:20.995504 (0.08039820156097412 s / img per device, on 2 devices)
2019-02-19 15:17:39,813 maskrcnn_benchmark.inference INFO: Preparing results for COCO format
2019-02-19 15:17:39,813 maskrcnn_benchmark.inference INFO: Preparing bbox results
2019-02-19 15:17:40,905 maskrcnn_benchmark.inference INFO: Preparing segm results
5000it [04:17, 19.11it/s]
2019-02-19 15:21:58,270 maskrcnn_benchmark.inference INFO: Evaluating predictions
Loading and preparing results...
DONE (t=1.59s)
creating index...
index created!
Running per image evaluation...
Evaluate annotation type *bbox*
DONE (t=32.39s).
Accumulating evaluation results...
DONE (t=5.60s).
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.060
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.151
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.034
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.039
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.077
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.070
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.090
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.174
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.188
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.087
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.197
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.224
Loading and preparing results...
DONE (t=4.55s)
creating index...
index created!
Running per image evaluation...
Evaluate annotation type *segm*
DONE (t=37.75s).
Accumulating evaluation results...
DONE (t=5.50s).
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.059
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.138
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.044
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.027
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.072
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.088
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.092
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.160
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.170
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.073
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.180
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.216
```
It is quite below the expected performance, and I now rerun the training to see if it is still the case.

One thing that was more odd, that the TQDM progress bar was counting 2500 not 5000 (while there are 5k annot in the `minival2014` set if I am right), and it was flickering awkwardly too much - nothing that is apparent when training / testing using polygons.

botcs:
Same issue on second run as well, box AP is 6.0 and mask AP is 5.9

JoyHuYY1412:
> @IssamLaradji I think it should be `w, h = box[2] - box[0] + 1, box[3] - box[1] + 1`.
> 
> And I guess `round` is more suitable than `int`?

In the code, should be round(float(b)) ?

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -1,51 +1,92 @@
...
+        flip_idx = torch.tensor(list(range(max_idx)[::-1]))
+        flipped_mask = self.mask.index_select(dim, flip_idx)
+        return Mask(flipped_mask, self.size, self.mode)
 
     def crop(self, box):
-        w, h = box[2] - box[0], box[3] - box[1]
-
-        cropped_masks = self.masks[:, box[1] : box[3], box[0] : box[2]]
-        return Mask(cropped_masks, size=(w, h), mode=self.mode)
+        box = [round(b) for b in box]
```
JoyHuYY1412: `box=[round(float(b) for b in box]` instead? since the box is tensor.

botcs:
Hi guys,

A few days ago @fmassa mentioned in one of [his comments](https://github.com/facebookresearch/maskrcnn-benchmark/pull/150#discussion_r258041747) the following:
> In general, I think both `box_list` and `segmentation_mask` could benefit from some better design / cleanup, but I'm not sure what that would be

So I tried to reshape things a bit to accommodate better the different requirements, but it has radically changed a few concepts. I would be keen on to learn about your opinions on it: #473

maskrcnn_benchmark/structures/segmentation_mask.py
fmassa: Those are all reasonable points, and I'm willing to accept PRs that improve the overall consistency and software design of the codebase

botcs: @fmassa Thanks, these points were considered for the refactored version, PR #473

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: that is correct

IssamLaradji:
I am getting this with this code,
```
  File "/maskrcnn-benchmark/maskrcnn_benchmark/modeling/roi_heads/mask_head/loss.py", line 39, in project_masks_on_boxes 
    scaled_mask = cropped_mask.resize((M, M)) 
  File "/mnt/home/issam/Research_Ground/domain_adaptation/ann_utils.py", line 282, in resize 
    self.mask[None, None, :, :], (height, width), mode="bilinear" 
  File "/miniconda/envs/py36/lib/python3.6/site-packages/torch/nn/functional.py", line 2447, in interpolate 
    return torch._C._nn.upsample_bilinear2d(input, _output_size(2), align_corners) 
RuntimeError: invalid argument 2: input and output sizes should be greater than 0, but got input (H: 2, W: 0) output (H: 28, W: 28) at /pytorch/aten/src/THNN/generic/SpatialUpSamplingBilinear.c:19 
Uncaught exception. Entering post mortem debugging 
Running 'cont' or 'step' will restart the program 

```

`self.mask = tensor([], size=(2, 0))`

fmassa:
@IssamLaradji does this also happen with #473 ?

hujh14:
@IssamLaradji I think it should be `box = [max(round(float(b)), 0) for b in box]`. The bad input size comes from a crop when a box coordinate rounds to -1.

fmassa:
Thanks for the initial work on this PR @wangg12 !

IssamLaradji:
Does this work yet for RLE?  I am getting when I pass a list of RLEs.

```
File "/maskrcnn-benchmark/maskrcnn_benchmark/structures/segmentation_mask.py", line 73, in __init__
    if len(masks.shape) == 2:
AttributeError: 'list' object has no attribute 'shape'
```

ShihuaiXu:
AttributeError: 'list' object has no attribute 'shape'
I met the same problem!

botcs:
Hi @ShihuaiXu ,
Please visit the updates [here](https://github.com/facebookresearch/maskrcnn-benchmark/pull/473#discussion_r273775714)

PR #160, opened 2018-11-15
minor fixes for typo and C4 network

lanpa:
the code path goes wrong if SIZE_DIVISIBILITY is not assigned.

### DIFF
--- a/configs/caffe2/e2e_mask_rcnn_R_50_C4_1x_caffe2.yaml
+++ b/configs/caffe2/e2e_mask_rcnn_R_50_C4_1x_caffe2.yaml
@@ -7,3 +7,5 @@ MODEL:
   MASK_ON: True
 DATASETS:
   TEST: ("coco_2014_minival",)
+DATALOADER:
+  SIZE_DIVISIBILITY: 32
--- a/configs/caffe2/e2e_faster_rcnn_R_50_C4_1x_caffe2.yaml
+++ b/configs/caffe2/e2e_faster_rcnn_R_50_C4_1x_caffe2.yaml
@@ -3,3 +3,5 @@ MODEL:
   WEIGHT: "catalog://Caffe2Detectron/COCO/35857197/e2e_faster_rcnn_R-50-C4_1x"
 DATASETS:
   TEST: ("coco_2014_minival",)
+DATALOADER:
+  SIZE_DIVISIBILITY: 32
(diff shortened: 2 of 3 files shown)


configs/caffe2/e2e_mask_rcnn_R_50_C4_1x_caffe2.yaml
```diff
@@ -7,3 +7,5 @@ MODEL:
   MASK_ON: True
 DATASETS:
   TEST: ("coco_2014_minival",)
+DATALOADER:
+  SIZE_DIVISIBILITY: 32
```
fmassa: This is not needed, right?
This size-divisibility is only needed (for now) for FPN-based networks, because they upsample the feature maps by a fixed scale.

lanpa:
```
Traceback (most recent call last):
  File "webcam.py", line 80, in <module>
    main()
  File "webcam.py", line 71, in main
    composite = coco_demo.run_on_opencv_image(img)
  File "/home/dexter/git/maskrcnn-benchmark/demo/predictor.py", line 169, in run_on_opencv_image
    predictions = self.compute_prediction(image)
  File "/home/dexter/git/maskrcnn-benchmark/demo/predictor.py", line 196, in compute_prediction
    image_list = to_image_list(image, self.cfg.DATALOADER.SIZE_DIVISIBILITY)
  File "/home/dexter/anaconda3/lib/python3.6/site-packages/maskrcnn_benchmark-0.1-py3.6-linux-x86_64.egg/maskrcnn_benchmark/structures/image_list.py", line 44, in to_image_list
    assert tensors.dim() == 4
AssertionError
```

The root cause lies in `to_image_list()`, which code path depends on whether `size_divisible>0`. Modifying the yaml file is just a quick fix to get it running.

fmassa: I think your error is orthogonal to the patch.

You might want to unsqueeze a new dimension in your input tensor, or pass it as a `List[Tensor]` instead.

One question: do you have this error in the standard code or is it a modification of the code provided?

fmassa:
I don't think this is the right fix for the issue you were facing.
Can you send a PR with the typo fixes?

Thanks!

configs/caffe2/e2e_mask_rcnn_R_50_C4_1x_caffe2.yaml
lanpa: yap, this PR didn't fix the root cause. BTW, can you run the C4 net work without modification?

PR #162, opened 2018-11-16
Docker hotfix

miguelvr:
The argument names where being overridden by the CUDA image env variables, resulting in a failure in the NCCL2 installation.

This fixes the issue


miguelvr:
removed the explicit install of NCCL2, as it was creating problems and it comes installed in the nvidia/cuda image

PR #163, opened 2018-11-16
Adds tensorboardx support

### LINKED ISSUE #76: can add tensorboardx for visualize the training status?
## 🚀 Feature
current training information only display in console, this is not very anschauung，can add tensorboardx to support loss or feature map 's visualize

miguelvr:
Closes #76 

**Details:**

Logs all the losses into tensoarboard

**Requirements:**

Requires installation of tensorboardX
```
pip install tensorboardX
```

**Usage:** 
```
python tools/train_net.py --use-tensorboard --config-file /path/to/config.yml
```

### DIFF
--- a/maskrcnn_benchmark/engine/trainer.py
+++ b/maskrcnn_benchmark/engine/trainer.py
@@ -44,6 +44,7 @@ def do_train(
     device,
     checkpoint_period,
     arguments,
+    tb_logger=None
 ):
     logger = logging.getLogger("maskrcnn_benchmark.trainer")
     logger.info("Start training")
@@ -71,6 +72,11 @@ def do_train(
         loss_dict_reduced = reduce_loss_dict(loss_dict)
         losses_reduced = sum(loss for loss in loss_dict_reduced.values())
         meters.update(loss=losses_reduced, **loss_dict_reduced)
+
+        if tb_logger:
+            tb_logger.add_scalar('loss', losses_reduced, iteration)
+            for key, value in loss_dict_reduced.items():
+                tb_logger.add_scalar(key, value, iteration)
 
         optimizer.zero_grad()
         losses.backward()
--- a/tools/train_net.py
+++ b/tools/train_net.py
@@ -5,19 +5,21 @@ Basic training script for PyTorch
 
 # Set up custom environment before nearly anything else is imported
 # NOTE: this should be the first import (no not reorder)
-from maskrcnn_benchmark.utils.env import setup_environment  # noqa F401 isort:skip
 
 import argparse
 import os
+import time
+from datetime import datetime
 
 import torch
+
 from maskrcnn_benchmark.config import cfg
 from maskrcnn_benchmark.data import make_data_loader
-from maskrcnn_benchmark.solver import make_lr_scheduler
-from maskrcnn_benchmark.solver import make_optimizer
 from maskrcnn_benchmark.engine.inference import inference
 from maskrcnn_benchmark.engine.trainer import do_train
 from maskrcnn_benchmark.modeling.detector import build_detection_model
+from maskrcnn_benchmark.solver import make_lr_scheduler
+from maskrcnn_benchmark.solver import make_optimizer
 from maskrcnn_benchmark.utils.checkpoint import DetectronCheckpointer
 from maskrcnn_benchmark.utils.collect_env import collect_env_info
 from maskrcnn_benchmark.utils.comm import synchronize, get_rank
@@ -26,7 +28,7 @@ from maskrcnn_benchmark.utils.logger import setup_logger
 from maskrcnn_benchmark.utils.miscellaneous import mkdir
 
 
-def train(cfg, local_rank, distributed):
+def train(cfg, local_rank, distributed, use_tensorboard=False):
     model = build_detection_model(cfg)
     device = torch.device(cfg.MODEL.DEVICE)
     model.to(device)
@@ -40,6 +42,18 @@ def train(cfg, local_rank, distributed):
             # this should be removed if we update BatchNorm stats
             broadcast_buffers=False,
         )
+
+    tb_logger = None
+    if use_tensorboard:
+        try:
+            from tensorboardX import SummaryWriter
+            timestamp = datetime.fromtimestamp(time.time()).strftime('%Y%m%d-%H:%M')
+            tb_logger = SummaryWriter('logs/maskrcnn-{}'.format(timestamp))
+        except ImportError:
+            raise ImportError(
+                'To use tensorboard please install tensorboardX '
+                '[ pip install tensorflow tensorboardX ].'
+            )
 
     arguments = {}
     arguments["iteration"] = 0
@@ -71,6 +85,7 @@ def train(cfg, local_rank, distributed):
         device,
         checkpoint_period,
         arguments,
+        tb_logger=tb_logger
     )
 
     return model
@@ -122,6 +137,13 @@ def main():
         action="store_true",
     )
     parser.add_argument(
+        "--use-tensorboard",
+        dest="use_tensorboard",
+        help="Use tensorboardX logger (Requires tensorboardX installed)",
+        action="store_true",
+        default=False
+    )
+    parser.add_argument(
         "opts",
         help="Modify config options using the command-line",
         default=None,
@@ -160,7 +182,12 @@ def main():
         logger.info(config_str)
     logger.info("Running with config:\n{}".format(cfg))
 
-    model = train(cfg, args.local_rank, args.distributed)
+    model = train(
+        cfg=cfg,
+        local_rank=args.local_rank,
+        distributed=args.distributed,
+        use_tensorboard=args.use_tensorboard
+    )
 
     if not args.skip_test:
         test(cfg, model, args.distributed)


tools/train_net.py
```diff
@@ -5,19 +5,21 @@
 
 # Set up custom environment before nearly anything else is imported
 # NOTE: this should be the first import (no not reorder)
-from maskrcnn_benchmark.utils.env import setup_environment  # noqa F401 isort:skip
```
fmassa: Can you please keep this?

maskrcnn_benchmark/engine/trainer.py
```diff
@@ -72,6 +73,11 @@ def do_train(
         losses_reduced = sum(loss for loss in loss_dict_reduced.values())
         meters.update(loss=losses_reduced, **loss_dict_reduced)
 
+        if tb_logger:
```
fmassa: What about adding support for TensorBoard directly in the [`MetricLogger`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/utils/metric_logger.py)?
We could pass the `meters` as an argument to `do_train`, and it can be setup to use TensorBoard or not in the `train_net` file.

I was thinking about something like
```python
class MetricLogger(object):
    def __init__(self, delimiter="\t", summary_writer=None):
        self.meters = defaultdict(SmoothedValue)
        self.delimiter = delimiter
        self.summary_writer = summary_writer

    def update(self, **kwargs):
        for k, v in kwargs.items():
            if isinstance(v, torch.Tensor):
                v = v.item()
            assert isinstance(v, (float, int))
            self.meters[k].update(v)
             if self.summary_writer:
                 tb_logger.add_scalar(k, v, iteration)
...
```

We would need to have a reliable way of getting the `iteration`, but that shouldn't be too bad / hard.
This way we don't need to change almost anything in the training code.

What do you think?

miguelvr:
@fmassa what doesn't play well exactly? my implementation or your suggestion?

I already ran it with 4x GPUs and it seemed to work. But I agree that we should put it only in the master process...

miguelvr:
> It doesn't play well because each process could be potentially writing to the same file, and we would be storing nGPU time more data than necessary.

I'm already working on a solution :+1:

miguelvr:
please check if this seems good to you. I can still add the writer creation to the metric logger, but it seemed better to do so in a more explicit way.

maskrcnn_benchmark/utils/tensorboard.py
```diff
@@ -0,0 +1,19 @@
...
+def get_tensorboard_writer(local_rank, distributed):
+    try:
+        from tensorboardX import SummaryWriter
+    except ImportError:
+        raise ImportError(
+            'To use tensorboard please install tensorboardX '
+            '[ pip install tensorflow tensorboardX ].'
+        )
+
+    if not distributed or (distributed and local_rank == 0):
```
fmassa: Can't you simplify this condition to something like
```
if distributed_rank == 0:
    ...
```
?

Also, it's safer to use the global `distributed_rank` instead of `local_rank`, as `local_rank` can be zero for each machine.

miguelvr: I didn't know that variable existed 👍

miguelvr:
will do!

maskrcnn_benchmark/engine/trainer.py
miguelvr: iteration doesn't exist there.

would this be okay for you then?

```python
class MetricLogger(object):
    def __init__(self, delimiter="\t", summary_writer=None):
        self.meters = defaultdict(SmoothedValue)
        self.delimiter = delimiter
        self.summary_writer = summary_writer

    def update(self, **kwargs):
        for k, v in kwargs.items():
            if isinstance(v, torch.Tensor):
                v = v.item()
            assert isinstance(v, (float, int))
            self.meters[k].update(v)
            if self.summary_writer:
                self.summary_writer.add_scalar(k, v, self.meters[k].count)
```

fmassa: I actually think that I'd prefer another class for the TFLogger, as I mentioned before.

About the `.count`, this will work fine as long as the user doesn't want to restart training, in which case the iteration will be off.

miguelvr:
@fmassa just to be sure:

you want a `TensorboardLogger` that subclasses `MetricLogger` and does both the metric logging and tensorboard logging or a separate `TensorboardLogger` that just logs to tensorboard?

fmassa:
I was thinking about a `TensorBoardLogger` that does both metric logging and tensorboard logging. And it can be a drop-in replacement for the `MetricLogger`.

maskrcnn_benchmark/utils/tensorboard.py
zimenglan-sysu-512: is it better to add `output_dir` as parameter to `get_tensorboard_writer` func?

miguelvr: yes, I'm doing that already. will push it soon

miguelvr:
@fmassa what do you think of the recent changes?

maskrcnn_benchmark/config/defaults.py
```diff
@@ -272,3 +272,6 @@
 _C.OUTPUT_DIR = "."
 
 _C.PATHS_CATALOG = os.path.join(os.path.dirname(__file__), "paths_catalog.py")
+
+_C.TENSORBOARD_LOGDIR = "logs"
+_C.TENSORBOARD_EXP_NAME = "maskrcnn-benchmark"  # tensorboard experiment name
```
fmassa: Do we actually need those two config values? Can't we instead just use `OUTPUT_DIR` for the TensorBoard logger?

maskrcnn_benchmark/engine/trainer.py
```diff
@@ -44,10 +45,19 @@ def do_train(
     device,
     checkpoint_period,
     arguments,
+    tb_log_dir,
```
fmassa: What about passing the `meter` as an argument to the `do_train`? This way, we can define it in the `tools/train.py` file.

maskrcnn_benchmark/utils/checkpoint.py
```diff
@@ -36,8 +36,7 @@ def save(self, name, **kwargs):
         if not self.save_to_disk:
             return
 
-        data = {}
-        data["model"] = self.model.state_dict()
+        data = {"model": self.model.state_dict()}
```
fmassa: Can you revert this change? It's not really necessary

tools/train_net.py
```diff
@@ -41,9 +42,7 @@ def train(cfg, local_rank, distributed):
             broadcast_buffers=False,
         )
 
-    arguments = {}
-    arguments["iteration"] = 0
-
+    arguments = {"iteration": 0}
```
fmassa: Can you revert this change?

maskrcnn_benchmark/utils/metric_logger.py
```diff
@@ -61,3 +65,42 @@ def __str__(self):
                 "{}: {:.4f} ({:.4f})".format(name, meter.median, meter.global_avg)
             )
         return self.delimiter.join(loss_str)
+
+
+class TensorboardLogger(MetricLogger):
+    def __init__(self,
+                 log_dir='logs',
+                 exp_name='maskrcnn-benchmark',
```
fmassa: Can we pass a single argument here, instead of `log_dir` and `exp_name`, pass and `save_file` or something?
There is no real value in having both, given that they are immediately concatenated into a single string.

achalddave:
Hey, is there any progress on this? It looks like this is almost ready to merge. I've addressed the final few comments from @fmassa here: https://github.com/achalddave/maskrcnn-benchmark/tree/tensorboard ([changes](https://github.com/achalddave/maskrcnn-benchmark/compare/a2ad9ab..4210b77)). Should I create a new pull request?

miguelvr:
Hey, I had forgotten about this. I can finish the PR on Monday

maskrcnn_benchmark/utils/metric_logger.py
```diff
@@ -61,3 +65,42 @@ def __str__(self):
...
+
+    def update(self, ** kwargs):
+        super(TensorboardLogger, self).update(**kwargs)
+        if self.writer:
+            for k, v in kwargs.items():
+                if isinstance(v, torch.Tensor):
+                    v = v.item()
+                assert isinstance(v, (float, int))
+                self.writer.add_scalar(k, v, self.iteration)
+            self.iteration += 1
```
achalddave: I think this iteration update needs to be elsewhere, explicitly, since `do_train` calls `update` twice in the same iteration ([loss update]https://github.com/facebookresearch/maskrcnn-benchmark/pull/163/files#diff-29486803add8a1cde2a6e5b741434c7cR83), [time update](https://github.com/facebookresearch/maskrcnn-benchmark/pull/163/files#diff-29486803add8a1cde2a6e5b741434c7cR91)). Or the semantics of Meters can be changed so that update should only be called once per iteration?

fmassa:
@achalddave maybe you can send a PR to @miguelvr branch, and then he can merge this and I can merge the PR?

miguelvr:
That would be the easiest. I have been having some busy days, sorry about that

miguelvr:
@fmassa feel free to review again

maskrcnn_benchmark/engine/trainer.py
```diff
@@ -70,15 +70,18 @@ def do_train(
         # reduce losses over all GPUs for logging purposes
         loss_dict_reduced = reduce_loss_dict(loss_dict)
         losses_reduced = sum(loss for loss in loss_dict_reduced.values())
-        meters.update(loss=losses_reduced, **loss_dict_reduced)
```
fmassa: Can we avoid this change? It could affect the reported runtime due to the async nature of CUDA operations.

maskrcnn_benchmark/utils/metric_logger.py
fmassa: I still think that there is no need for two arguments: if you run two trainings on the same directory, it will have conflicts with the checkpoints and it will not work as expected.

In this case, I would just pass the full path to the file as an argument.

maskrcnn_benchmark/config/defaults.py
```diff
@@ -272,3 +272,5 @@
 _C.OUTPUT_DIR = "."
 
 _C.PATHS_CATALOG = os.path.join(os.path.dirname(__file__), "paths_catalog.py")
+
+_C.TENSORBOARD_EXP_NAME = "maskrcnn-benchmark"  # tensorboard experiment name
```
fmassa: I still think that there is no need for two arguments: if you run two trainings on the same directory, it will have conflicts with the checkpoints and it will not work as expected.

In this case, I would just pass the same `OUTPUT_DIR` here.

miguelvr:
Sure, I can address those!

maskrcnn_benchmark/config/defaults.py
miguelvr: It will not conflict because I add a time stamp. 

What is your idea? To have each tensorboard log in the model directory instead of a dedicated logs folder?

fmassa: What I meant is that if you try to run multiple training experiments in the same `OUTPUT_DIR` folder, then you'll have problems during training, as the `Checkpointer` might pick the checkpoint from a different experiment.

If you would want to have all the TensorBoard logs in a single place (because it allows to compare different experiments), then I'd keep an extra argument in the config (or the training script) with the full path, and you can add the timestamp in your logic.

maskrcnn_benchmark/utils/metric_logger.py
```diff
@@ -61,3 +65,41 @@ def __str__(self):
                 "{}: {:.4f} ({:.4f})".format(name, meter.median, meter.global_avg)
             )
         return self.delimiter.join(loss_str)
+
+
+class TensorboardLogger(MetricLogger):
+    def __init__(self,
+                 log_dir='logs/maskrcnn-benchmark',
```
fmassa: can we avoid having a default argument for this value? I think it's better to keep a default argument in the place where it's called

maskrcnn_benchmark/config/defaults.py
miguelvr: that sounds better!

maskrcnn_benchmark/utils/metric_logger.py
miguelvr: sure, on it

maskrcnn_benchmark/engine/trainer.py
fmassa: Just this change left, and then I believe this looks good to be merged!

miguelvr: i just put it back

maskrcnn_benchmark/utils/metric_logger.py
miguelvr: yes, this needs to be addressed

miguelvr: it should be fine, we are looping on kwargs, so only on the keys that should be updated.

miguelvr:
@fmassa everything should be solved now

maskrcnn_benchmark/engine/trainer.py
achalddave: Keeping the two updates separate will still cause the iteration count to be off by a factor of 2, right? Should we explicitly define and call meters.update_iter() or similar to only update the iteration count once per loop?

maskrcnn_benchmark/config/defaults.py
achalddave: Can't the tensorboard logs all be in the OUTPUT_DIR instead of a separate logs directory, but still be compared? This is how the unofficial pytorch detectron [handles it](https://github.com/roytseng-tw/Detectron.pytorch/blob/9defe952cb8b1a4b90b1c1ad19eb5b3614b01418/tools/train_net.py#L301). As long as all the model directories are in the same root directory, e.g.
`experiments/model1/`, `experiments/model2/`, `experiments/model3/`, each containing tensorboard summaries, pointing tensorboard to the `experiments` directory will show all the logs.

miguelvr: admitting that every model has the same root output dir, yes.

JoyHuYY1412:
> @fmassa what doesn't play well exactly? my implementation or your suggestion?
> 
> I already ran it with 4x GPUs and it seemed to work. But I agree that we should put it only in the master process...

Thank you for your work. I tried to use multi-gpu like `python -m torch.distributed.launch --nproc_per_node=2 tools/train_net.py --config-file "configs/caffe2/e2e_mask_rcnn_R_50_FPN_1x_caffe2.yaml" SOLVER.IMS_PER_BATCH 2 SOLVER.BASE_LR 0.01 SOLVER.MAX_ITER 720000 SOLVER.STEPS "(480000, 640000)" TEST.IMS_PER_BATCH 1 OUTPUT_DIR "./ckp_sm
antic"`
but the code just stopped before training. After I delete --use-tensorboard, it runs without problem. Could you give me some advice?

fmassa:
@miguelvr have you tried using multiple GPUs? I don't see anything wrong with the current implementation, but @JoyHuYY1412 is reporting problems with this PR

miguelvr:
I have access multiple GPUs, just not available at the moment. I will check this out ASAP

achalddave:
Hm.. I've been training with the code from [a slightly earlier version](https://github.com/facebookresearch/maskrcnn-benchmark/pull/163/commits/0c2696b6c96582a209075fd43d16b7ee28a4335a) with multiple GPUs for a few days now. I can try the most recent version soon as well. @JoyHuYY1412 do you receive any particular error? If the code is simply hanging, is it possible for you to pinpoint where the code hangs?

Iwontbecreative:
Is this still active?

miguelvr:
@fmassa just merged master into this and tried it out with 2 GPUs and it worked fine!

feel free to review again

miguelvr:
pinging this

miguelvr:
@fmassa what's missing for this to be merged?

maxsenh:
Hi this is really interesting! I have been trying to implement this in my workflow. How do I call tensorboardx? With `tensorboard --logdir runs` ?

miguelvr:
Hi @maxsenh,

Can you check that the cfg.TENSORBOARD_EXPERIMENT path contains files after training?

It has been working for me with no problems

maskrcnn_benchmark/config/defaults.py
```diff
@@ -421,3 +421,5 @@
 _C.OUTPUT_DIR = "."
 
 _C.PATHS_CATALOG = os.path.join(os.path.dirname(__file__), "paths_catalog.py")
+
+_C.TENSORBOARD_EXPERIMENT = "logs/maskrcnn-benchmark"  # tensorboard experiment location
```
rodrigoberriel: nit: if you don't mind the suggestion, all the other comments are above :smile: 

```suggestion
# TensorBoard experiment location
_C.TENSORBOARD_EXPERIMENT = "logs/maskrcnn-benchmark"
```

tools/train_net.py
```diff
@@ -61,6 +64,13 @@ def train(cfg, local_rank, distributed):
     )
 
     checkpoint_period = cfg.SOLVER.CHECKPOINT_PERIOD
+    if use_tensorboard:
+        meters = TensorboardLogger(
+            log_dir=cfg.TENSORBOARD_EXPERIMENT,
```
rodrigoberriel: If you don't mind the question: why did you choose to use another directory instead of using something like `os.path.join(cfg.OUTPUT_DIR, "tensorboard")`?

tools/train_net.py
```diff
@@ -11,22 +11,25 @@
...
 from maskrcnn_benchmark.modeling.detector import build_detection_model
+from maskrcnn_benchmark.solver import make_lr_scheduler
+from maskrcnn_benchmark.solver import make_optimizer
 from maskrcnn_benchmark.utils.checkpoint import DetectronCheckpointer
 from maskrcnn_benchmark.utils.collect_env import collect_env_info
 from maskrcnn_benchmark.utils.comm import synchronize, get_rank
 from maskrcnn_benchmark.utils.imports import import_file
 from maskrcnn_benchmark.utils.logger import setup_logger
+from maskrcnn_benchmark.utils.metric_logger import (
+    MetricLogger, TensorboardLogger)
```
rodrigoberriel: nit: if we are going to use 2 lines, why not to use the same style used in the `solver` imports (L20-21)? Also, IMO, it would be nice if you could rename `TensorboardLogger` to `TensorBoardLogger` :laughing:

```python
from maskrcnn_benchmark.utils.metric_logger import MetricLogger
from maskrcnn_benchmark.utils.metric_logger import TensorBoardLogger
```

tools/train_net.py
miguelvr: because you may want to store all your tensorboard logs on the same directory.

tools/train_net.py
miguelvr: why would you separate imports from the same module?

maxsenh:
> Hi @maxsenh,
> 
> Can you check that the cfg.TENSORBOARD_EXPERIMENT path contains files after training?
> 
> It has been working for me with no problems

It is working now for me as well, thanks for the help! The issue was related to the fact that I am running this on a cluster. Do you have any idea on how to plot the validation loss as well?

miguelvr:
> > Hi @maxsenh,
> > Can you check that the cfg.TENSORBOARD_EXPERIMENT path contains files after training?
> > It has been working for me with no problems
> 
> It is working now for me as well, thanks for the help! The issue was related to the fact that I am running this on a cluster. Do you have any idea on how to plot the validation loss as well?

there is no validation in the repository yet, only testing on the last checkpoint.

I would like to add it once we have proper validation.

tools/train_net.py
rodrigoberriel: Sorry, I forgot to add a "as the default" to my question. I would expect that the default is to save all output data, including tensorboard-related files in the `OUTPUT_DIR`. Maybe this is only me, idk.

tools/train_net.py
rodrigoberriel: I wouldn't, but this would be the only exception in this repository, if I am not wrong. I have no problem with multi-line imports ([PEP 328](https://www.python.org/dev/peps/pep-0328/)), but I do with inconsistencies :smile:

miguelvr:
@fmassa should I update this with the new pytorch TensorBoard integration>

droseger:
@miguelvr Thanks, your contribution is very helpful!
There seems to be an issue with the iteration count, though. Depending on the stat, it displays either only even or uneven iteration numbers and thus the reported number of the last iteration is two times the actual number.

wangg12:
Hi all, when will this feature be merged?

botcs:
Hi @fmassa,
Any updates on this?

yangjzx:
Is this active?

miguelvr:
I have had no feedback from @fmassa in a long time, but I would be happy to see this merged

yangjzx:
Thanks for the help! I find a problem when using this implementation. When we using tsensorboardx, it seems not simple if we want to separately write something into default logger and not into tensorboard_write. For example, the batch_time and data_time will be helpful to know how much time left for the training process, but I will not want them presenting on the tensorboard web page.

maskrcnn_benchmark/utils/metric_logger.py
```diff
@@ -64,3 +68,41 @@ def __str__(self):
...
+
+    def update(self, ** kwargs):
+        super(TensorboardLogger, self).update(**kwargs)
+        if self.writer:
+            for k, v in kwargs.items():
+                if isinstance(v, torch.Tensor):
+                    v = v.item()
+                assert isinstance(v, (float, int))
+                self.writer.add_scalar(k, v, self.iteration)
+            self.iteration += 1
```
yangjzx: self.iteration += 1
will cause the iterations logged in tensorboard inconsistent with the training iterations. As in the trainer.py, there are two meters.update within one iteration. From my opinion, it may be better to get self.iteration from the arguments iteration or arguments["iteration"] in do_train() in trainer.py.

botcs: @miguelvr

TyroneLi:
Did anybody try to use add_graphy func? I tried but fail due to the args inputed to forward func were not tensors or some required datatype?

PR #164, opened 2018-11-16
Add returning encoding features of detected objects in RCNN

realEbi:
It is a good idea to maintain the indices of filtered bboxes for RCNN detection.  
One application is to return representation or encoded features of each bbox in inference time.  

# Description
Preparing encoded features of detected objects and storing them into an HDF5 file.
These features are needed for bottom-up attention features described at [this paper](https://arxiv.org/abs/1707.07998)

# Requirements
`pip install h5py`

# Usage
To extract and store encoded features from objects in images in `folder_path` execute the following command:  
```bash
python demo/roi_encoding.py
```
The function in `COCODemo.extract_encoding_features` extract encoded features from detected objects exists in the input images.

### DIFF
--- a/demo/predictor.py
+++ b/demo/predictor.py
@@ -8,6 +8,8 @@ from maskrcnn_benchmark.utils.checkpoint import DetectronCheckpointer
 from maskrcnn_benchmark.structures.image_list import to_image_list
 from maskrcnn_benchmark.modeling.roi_heads.mask_head.inference import Masker
 
+def hook(module, input, output):
+            setattr(module, "_value_hook", output)
 
 class COCODemo(object):
     # COCO categories for pretty print
@@ -108,6 +110,13 @@ class COCODemo(object):
         self.model.eval()
         self.device = torch.device(cfg.MODEL.DEVICE)
         self.model.to(self.device)
+        
+
+        for n, m in self.model.named_modules():
+            if n == "roi_heads":
+                m.register_forward_hook(hook)
+
+        
         self.min_image_size = min_image_size
 
         checkpointer = DetectronCheckpointer(cfg, self.model)
@@ -370,7 +379,10 @@ class COCODemo(object):
         image_list = image_list.to(self.device)
         # compute predictions
         with torch.no_grad():
-            predictions, features = self.model(image_list, return_features=True)
+            predictions = self.model(image_list)
+            for n, m in self.model.named_modules():
+                if n == "roi_heads":
+                    features = m._value_hook[0]
         predictions = [o.to(self.cpu_device) for o in predictions]
         features = [o.to(self.cpu_device).unsqueeze(0) for o in features]
         detection_bboxes=[]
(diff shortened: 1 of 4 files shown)


maskrcnn_benchmark/modeling/detector/generalized_rcnn.py
```diff
@@ -61,5 +64,7 @@ def forward(self, images, targets=None):
             losses.update(detector_losses)
             losses.update(proposal_losses)
             return losses
-
-        return result
+        if return_features:
```
fmassa: Instead of hard-coding the last features, what do you think about the user attaching a hook in their model to obtain the output of a particular layer? This would be a non-intrusive change to this file, and the feature that the user could obtain can be arbitrary

demo/predictor.py
```diff
@@ -353,3 +353,38 @@ def overlay_class_names(self, image, predictions):
...
+        features = [o.to(self.cpu_device).unsqueeze(0) for o in features]
+        detection_bboxes=[]
+        for ind, prediction in enumerate(predictions):
+            height, width = original_images[ind].shape[:-1]
+            # reshape prediction (a BoxList) into the original image size
+            prediction = prediction.resize((width, height))
+            detection_bbox = self.select_top_predictions(prediction)
+
+            if len(detection_bbox.extra_fields['orig_inds'])>0:
+                encoding_features = torch.cat([features[1000*ind+i] for i in detection_bbox.extra_fields['orig_inds']], dim=0)
```
fmassa: I'm a bit confused. What do the `1000` represent? Is it the max number of proposals that is fed to the classifier?

fmassa (commented):
Hi,

Thanks for the PR.

I have a few questions so that I understand a bit better this PR.

maskrcnn_benchmark/modeling/detector/generalized_rcnn.py
e3oroush: That would be much more logical. But I couldn't come up with a better approach at the time. Adding [hook](https://pytorch.org/docs/stable/nn.html#torch.nn.Module.register_forward_hook) didn't come to my mind. I'll amend the PR with this method.

TheShadow29:
@e3oroush I am trying to use your code to extract features from mask rcnn model. I noticed that in `predictor.py` you use `max proposals`.

`max_proposals` are set to `self.cfg.MODEL.RPN.FPN_POST_NMS_TOP_N_TEST`

 I am unable to understand that logic. The `features` output I am getting has the same number as the `predictions`, for example, in my particular case, I get 56 boxes in predictions and features is of size torch.Size([56, 256, 14, 14]).

Can you suggest what I am missing?

e3oroush:
@TheShadow29 Sorry for this late response.
I use it for object detection rcnn I didn't use it for maskrcnn.
`max_proposal` represnts maximum number of object proposals that are fed to the classifier.
You could use a `max_pooling` to convert [256,14,14] -> [1024] for each of mask boxes.

HuaizhengZhang:
@e3oroush Hi, thanks for your great PR. This is really helpful. Have you tried to compare your results to results from this code? https://github.com/facebookresearch/pythia/blob/master/tools/extract_features.py

e3oroush:
> @e3oroush Hi, thanks for your great PR. This is really helpful. Have you tried to compare your results to results from this code? https://github.com/facebookresearch/pythia/blob/master/tools/extract_features.py

Thanks for your kind comment. Actually, no I didn't. I'll compare them and let you know as soon as I can.

botcs:
Hey @e3oroush @TheShadow29,

> Thanks for your kind comment. Actually, no I didn't. I'll compare them and let you know as soon as I can.

Can you please update this PR?

PR #179, opened 2018-11-19
Add posibility to setup DATA_DIR as env variable

vfdev-5:
Hi Francisco,

What do you think about such tweak to have a possibilty to setup DATA_DIR from the environment variable instead of symlinking folders into `datasets` as described in the README ?

### DIFF
--- a/maskrcnn_benchmark/config/paths_catalog.py
+++ b/maskrcnn_benchmark/config/paths_catalog.py
@@ -5,7 +5,7 @@ import os
 
 
 class DatasetCatalog(object):
-    DATA_DIR = "datasets"
+    DATA_DIR = "datasets" if "DATA_DIR" not in os.environ else os.environ["DATA_DIR"]
 
     DATASETS = {
         "coco_2014_train": (


fmassa:
Hey Victor,

While I'm fine with this change, it introduces an extra level of complexity to the user.
What if for example I want to change the folder for COCO, but not for Pascal dataset? Should I now allow a few more different env vars?

The way I'm currently advertising doing things if there is a substantial change is to pass their own `paths_catalog.py` file, which gives you total flexibility.
For example
```python
from maskrcnn_benchmark.config.paths_catalog import ModelCatalog, DatasetCatalog
DatasetCatalog.DATA_DIR = "/my/path"
```
and it also gives you much more flexibility.

Creating a symlink to the folders is also pretty easy, so I'd maybe stick with those two approaches that are already currently possible, and not maybe adding more complexity.

Thoughts?

vfdev-5:
No problem, actually, I did not realized that `DATA_DIR` is such generic path for MSCoco and probably future VOC. But OK, your proposition with 
```
from maskrcnn_benchmark.config.paths_catalog import ModelCatalog, DatasetCatalog
DatasetCatalog.DATA_DIR = "/my/path"
```
seems OK to me. Feel free to close this one.

fmassa:
Cool, thanks for the PR anyway!

PR #183, opened 2018-11-20
add opencv support to Dockerfile

keineahnung2345:
cv2 is used by predictor.py, so it's required

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -7,9 +7,7 @@ RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selectio
 
 # install basics
 RUN apt-get update -y \
- && apt-get install -y apt-utils git curl ca-certificates bzip2 cmake tree htop bmon iotop g++ \
- && apt-get install -y libglib2.0-0 libsm6 libxext6 libxrender-dev
- 
+ && apt-get install -y apt-utils git curl ca-certificates bzip2 cmake tree htop bmon iotop g++
 
 # Install Miniconda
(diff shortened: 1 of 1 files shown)


miguelvr:
I would prefer just to keep it in conda, as it reduces the boilerplate and guarantees a correct installation

docker/Dockerfile
```diff
@@ -28,7 +30,7 @@ ENV PATH=$CONDA_PREFIX/bin:$PATH
 ENV CONDA_AUTO_UPDATE_CONDA=false
 
 RUN conda install -y ipython
-RUN pip install ninja yacs cython matplotlib
+RUN pip install ninja yacs cython matplotlib opencv-python
 
 # Install PyTorch 1.0 Nightly
 RUN conda install -y pytorch-nightly -c pytorch && conda clean -ya
```
miguelvr: Replace with

```
# Install PyTorch 1.0 Nightly and OpenCV
RUN conda install -y pytorch-nightly -c pytorch \
 && conda install -y opencv -c menpo \
 && conda clean -ya
```

and remove your added apt-get install (...)

keineahnung2345: I've tried testing the effect of installing opencv through conda, and it works without my apt-get install xxx, thanks!

keineahnung2345:
@miguelvr I have changed to conda. But I wonder why you install ninja, yacs, cython, matplotlib through pip?

fmassa:
@keineahnung2345 that's the default requirements for the installation for those libs, and is what we recommend in the installation.

miguelvr:
@keineahnung2345 basically I just put it as in the INSTALL.md for clarity.

The benefit of installing opencv through conda is just to avoid the clutter of all the dependencies it has

PR #196, opened 2018-11-21
Initial mixed-precision training

slayton58:
This PR adds initial mixed-precision training support via [apex.amp](https://github.com/NVIDIA/apex/tree/master/apex/amp).

Mixed-precision is controlled with the `SOLVER.MIXED_PRECISION` config argument.

Along with `apex.amp` support, I've moved `DistributedDataParallel` to `apex.DistributedDataParallel` as this is what we've been using to good effect over the last few months.

Please note that this does add the `apex` package as a requirement.

### DIFF
--- a/maskrcnn_benchmark/layers/batch_norm.py
+++ b/maskrcnn_benchmark/layers/batch_norm.py
@@ -18,7 +18,7 @@ class FrozenBatchNorm2d(nn.Module):
 
     def forward(self, x):
         # Cast all fixed parameters to half() if necessary
-        if x.type() == torch.half:
+        if x.dtype == torch.float16:
             self.weight = self.weight.half()
             self.bias = self.bias.half()
             self.running_mean = self.running_mean.half()
--- a/maskrcnn_benchmark/layers/batch_norm.py
+++ b/maskrcnn_benchmark/layers/batch_norm.py
@@ -17,6 +17,13 @@ class FrozenBatchNorm2d(nn.Module):
         self.register_buffer("running_var", torch.ones(n))
 
     def forward(self, x):
+        # Cast all fixed parameters to half() if necessary
+        if x.type() == torch.half:
+            self.weight = self.weight.half()
+            self.bias = self.bias.half()
+            self.running_mean = self.running_mean.half()
+            self.running_var = self.running_var.half()
+
         scale = self.weight * self.running_var.rsqrt()
         bias = self.bias - self.running_mean * scale
         scale = scale.reshape(1, -1, 1, 1)
--- a/maskrcnn_benchmark/modeling/poolers.py
+++ b/maskrcnn_benchmark/modeling/poolers.py
@@ -116,6 +116,6 @@ class Pooler(nn.Module):
         for level, (per_level_feature, pooler) in enumerate(zip(x, self.poolers)):
             idx_in_level = torch.nonzero(levels == level).squeeze(1)
             rois_per_level = rois[idx_in_level]
-            result[idx_in_level] = pooler(per_level_feature, rois_per_level)
+            result[idx_in_level] = pooler(per_level_feature, rois_per_level).to(dtype)
 
         return result
--- a/tools/train_net.py
+++ b/tools/train_net.py
@@ -31,7 +31,7 @@ try:
     from apex.parallel import DistributedDataParallel as DDP
     from apex import amp
 except ImportError:
-    print('Use APEX for better performance via apex.amp and apex.DistributedDataParallel')
+    raise ImportError('Use APEX for better performance via apex.amp and apex.DistributedDataParallel')
 
 
 def train(cfg, local_rank, distributed):
@@ -42,9 +42,11 @@ def train(cfg, local_rank, distributed):
     optimizer = make_optimizer(cfg, model)
     scheduler = make_lr_scheduler(cfg, optimizer)
 
-    # Wrap the optimizer for fp16 training
-    use_mixed_precision = cfg.SOLVER.MIXED_PRECISION
-    amp_handle = amp.init(enabled=use_mixed_precision, verbose=False)
+    # Initialize mixed-precision training
+    use_mixed_precision = cfg.DTYPE == "float16"
+    amp_handle = amp.init(enabled=use_mixed_precision, verbose=cfg.AMP_VERBOSE)
+
+    # wrap the optimizer for mixed precision
     optimizer = amp_handle.wrap_optimizer(optimizer)
 
     if distributed:
--- a/tools/train_net.py
+++ b/tools/train_net.py
@@ -25,6 +25,14 @@ from maskrcnn_benchmark.utils.imports import import_file
 from maskrcnn_benchmark.utils.logger import setup_logger
 from maskrcnn_benchmark.utils.miscellaneous import mkdir
 
+# See if we can use apex.DistributedDataParallel instead of the torch default,
+# and enable mixed-precision via apex.amp
+try:
+    from apex.parallel import DistributedDataParallel as DDP
+    from apex import amp
+except ImportError:
+    print('Use APEX for better performance via apex.amp and apex.DistributedDataParallel')
+
 
 def train(cfg, local_rank, distributed):
     model = build_detection_model(cfg)
@@ -33,6 +41,11 @@ def train(cfg, local_rank, distributed):
 
     optimizer = make_optimizer(cfg, model)
     scheduler = make_lr_scheduler(cfg, optimizer)
+
+    # Wrap the optimizer for fp16 training
+    use_mixed_precision = cfg.SOLVER.MIXED_PRECISION
+    amp_handle = amp.init(enabled=use_mixed_precision, verbose=False)
+    optimizer = amp_handle.wrap_optimizer(optimizer)
 
     if distributed:
         model = torch.nn.parallel.DistributedDataParallel(
--- a/tools/train_net.py
+++ b/tools/train_net.py
@@ -43,10 +43,8 @@ def train(cfg, local_rank, distributed):
 
     # Initialize mixed-precision training
     use_mixed_precision = cfg.DTYPE == "float16"
-    amp_handle = amp.init(enabled=use_mixed_precision, verbose=cfg.AMP_VERBOSE)
-
-    # wrap the optimizer for mixed precision
-    optimizer = amp_handle.wrap_optimizer(optimizer)
+    amp_opt_level = 'O1' if use_mixed_precision else 'O0'
+    model, optimizer = amp.initialize(model, optimizer, opt_level=amp_opt_level)
 
     if distributed:
         model = torch.nn.parallel.DistributedDataParallel(
--- a/tools/train_net.py
+++ b/tools/train_net.py
@@ -28,10 +28,9 @@ from maskrcnn_benchmark.utils.miscellaneous import mkdir
 # See if we can use apex.DistributedDataParallel instead of the torch default,
 # and enable mixed-precision via apex.amp
 try:
-    from apex.parallel import DistributedDataParallel as DDP
     from apex import amp
 except ImportError:
-    raise ImportError('Use APEX for better performance via apex.amp and apex.DistributedDataParallel')
+    raise ImportError('Use APEX for multi-precision via apex.amp')
 
 
 def train(cfg, local_rank, distributed):
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -42,6 +42,12 @@ git clone https://github.com/cocodataset/cocoapi.git
 cd cocoapi/PythonAPI
 python setup.py build_ext install
 
+# install apex
+cd ~github
+git clone https://github.com/NVIDIA/apex.git
+cd apex
+python setup.py install --cuda_ext --cpp_ext
+
 # install PyTorch Detection
 cd ~/github
 git clone https://github.com/facebookresearch/maskrcnn-benchmark.git
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -45,6 +45,11 @@ RUN git clone https://github.com/cocodataset/cocoapi.git \
  && cd cocoapi/PythonAPI \
  && python setup.py build_ext install
 
+# install apex
+RUN git clone https://github.com/NVIDIA/apex.git \
+ && cd apex \
+ && python setup.py install --cuda_ext --cpp_ext
+
 # install PyTorch Detection
 RUN git clone https://github.com/facebookresearch/maskrcnn-benchmark.git \
  && cd maskrcnn-benchmark \
--- a/maskrcnn_benchmark/config/defaults.py
+++ b/maskrcnn_benchmark/config/defaults.py
@@ -254,6 +254,9 @@ _C.SOLVER.CHECKPOINT_PERIOD = 2500
 # see 2 images per batch
 _C.SOLVER.IMS_PER_BATCH = 16
 
+# Whether or not to use mixed-precision (via apex.amp)
+_C.SOLVER.MIXED_PRECISION = False
+
 # ---------------------------------------------------------------------------- #
 # Specific test options
 # ---------------------------------------------------------------------------- #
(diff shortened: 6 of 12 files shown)


miguelvr:
could you please add it to the Dockerfile as well?

slayton58:
@miguelvr Done - hadn't noticed that file was there :)

maskrcnn_benchmark/layers/batch_norm.py
```diff
@@ -17,6 +17,13 @@ def __init__(self, n):
         self.register_buffer("running_var", torch.ones(n))
 
     def forward(self, x):
+        # Cast all fixed parameters to half() if necessary
+        if x.type() == torch.half:
+            self.weight = self.weight.half()
```
fmassa: So it seems that we don't explicitly cast the model to `fp16` during initialization, is that right?
This seems a bit counter-intuitive to me, what happens if we just cast everything in `model` to `.half()`?

maskrcnn_benchmark/layers/batch_norm.py
```diff
@@ -17,6 +17,13 @@ def __init__(self, n):
         self.register_buffer("running_var", torch.ones(n))
 
     def forward(self, x):
+        # Cast all fixed parameters to half() if necessary
+        if x.type() == torch.half:
```
fmassa: If we need to keep this, I think the preferred solution is to do `x.dtype == torch.float16`. (`x.type()` is deprecated or kind of )

maskrcnn_benchmark/modeling/poolers.py
```diff
@@ -116,6 +116,6 @@ def forward(self, x, boxes):
         for level, (per_level_feature, pooler) in enumerate(zip(x, self.poolers)):
             idx_in_level = torch.nonzero(levels == level).squeeze(1)
             rois_per_level = rois[idx_in_level]
-            result[idx_in_level] = pooler(per_level_feature, rois_per_level)
+            result[idx_in_level] = pooler(per_level_feature, rois_per_level).to(dtype)
```
fmassa: doesn't the `amp.float_function` wrap back the values to fp16 after they are computed?

tools/train_net.py
```diff
@@ -25,6 +25,14 @@
 from maskrcnn_benchmark.utils.logger import setup_logger
 from maskrcnn_benchmark.utils.miscellaneous import mkdir
 
+# See if we can use apex.DistributedDataParallel instead of the torch default,
+# and enable mixed-precision via apex.amp
+try:
+    from apex.parallel import DistributedDataParallel as DDP
+    from apex import amp
+except ImportError:
+    print('Use APEX for better performance via apex.amp and apex.DistributedDataParallel')
```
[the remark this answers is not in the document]
slayton58: What's your preferred way of doing this? `quit()`?

tools/train_net.py
```diff
@@ -34,11 +42,14 @@ def train(cfg, local_rank, distributed):
...
+    use_mixed_precision = cfg.SOLVER.MIXED_PRECISION
+    amp_handle = amp.init(enabled=use_mixed_precision, verbose=False)
+    optimizer = amp_handle.wrap_optimizer(optimizer)
+
     if distributed:
-        model = torch.nn.parallel.deprecated.DistributedDataParallel(
-            model, device_ids=[local_rank], output_device=local_rank,
-            # this should be removed if we update BatchNorm stats
-            broadcast_buffers=False,
+        model = DDP(
```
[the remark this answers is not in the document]
slayton58: `apex.DistributedDataParallel` has similar perf to the c10d implementation, and it's what we've been running for the last few months -- I'm not married to the change, but I might suggest using it until the c10d implementation works for Mask-RCNN (if it doesn't already in master)

maskrcnn_benchmark/modeling/poolers.py
slayton58: I'd need to run again to work out exactly what case was happening here, but the type change was not happening correctly and I had to manually cast here to prevent errors

tools/train_net.py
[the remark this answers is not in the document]
slayton58: We shouldn't, but I haven't tested it.

maskrcnn_benchmark/layers/batch_norm.py
slayton58: In using "support" I chose my words badly. It's not necessarily support, more "can be used with a reasonable expectation of not losing accuracy". Apex.amp takes a conservative approach by not moving ops to fp16 when we're not sure of their accuracy (the lists of what is / isn't moved to fp16 are in the files [here](https://github.com/NVIDIA/apex/tree/master/apex/amp/lists).

We could try casting the entire model to half and see what happens -- there's enough code in the RPN especially that I'm just not sure how it'll behave in lower precisions that I decided to be conservative beyond what apex.amp does. Unfortunately that means until PyT can grok `y_16 = a_32 * x_16 + b_32` where the subscripts denote precision we have to do something manual here

miguelvr:
any updates on this?

slayton58:
I'm rebasing right now (along with moving back to pytorch default `DistributedDataParallel`, otherwise Im waiting on more reviews / feedback

maskrcnn_benchmark/modeling/poolers.py
fmassa: Quick question: is this casting still relevant?

tools/train_net.py
```diff
@@ -25,6 +25,14 @@
 from maskrcnn_benchmark.utils.logger import setup_logger
 from maskrcnn_benchmark.utils.miscellaneous import mkdir
 
+# See if we can use apex.DistributedDataParallel instead of the torch default,
+# and enable mixed-precision via apex.amp
+try:
+    from apex.parallel import DistributedDataParallel as DDP
```
fmassa: This import is now not necessary I think

maskrcnn_benchmark/modeling/poolers.py
slayton58: Yes: `result[idx_in_level]` is expected to be in fp16 (as it's the same precision as the input), but the pooler returns fp32 (explicitly, as the code hasn't had fp16 support added). To get around this the result from the pooler needs to be casted

fmassa: But I thought that `amp.float_function` would:
1 - cast to float
2 - compute
3 - cast back to fp16

Or is my understanding of it wrong?

slayton58: By my understanding this is an artifact of how MRCNN is constructed. The input to Pooler is cast to `half`. It initialises the output for pooling to that type. Then the Pooler module calls the `ROIPool` module, which casts _it's_ input to `float`. It doesn't explicitly cast back to `half` (the next module in line would do this for you), so it tries to assign a `float` tensor to a `half` result. This only occurs because the poolers are members of a Module, rather than composed with that module using `nn.Sequential` or otherwise

fmassa: Ok.
I think the solution I'd potentially do myself (while support for fp16 is not present in the core pooler functions) is to just cast in the C++ side to float and cast back if the type is fp16.
But I suppose this is not really a hard-requirement here (but would make things cleaner)

slayton58: Would you like a comment added explaining the current need for the cast? (along with a TODO for the full (fp16 support in pooling) if you so desire)

slayton58: Because this is a strange case -- you're in fp16-land, allocate your output as fp16, then run something that has to cast up to fp32. There's no cast back automatically (you're calling from a module, so there's no module boundary to cause a cast) so it has to be done manually. If lines 111-115 didn't exist, and you inferred the type from the return type of the pooling call this explicit wouldn't have to be there.

slayton58: I understand it's a little weird, but that's where the code is right now -- if you want an explicit (non-AMP) fp16 version that can also be done, but it'll be more invasive and can't be done before the new year after I get back

wat3rBro:
@slayton58 Hi, I'm collecting stats of training, finding out it would terminate silently during the first iteration if fp16 is enabled for training ResNeXt model (both detection and mask) on V100. I checked the verbose output from APEX and couldn't find difference when comparing it with a successful run on P100 for the same model. The exit should happen when running:
```
with optimizer.scale_loss(losses) as scaled_losses:
    scaled_losses.backward()
```
Is this a known issue?

slayton58:
@wat3rBro I've never run with ResNeXt (it didn't exist on the version of the code I originally developed this for). There should be no difference in behavior between resnet and resnext, does this happen on the first iteration? Also, do other tests pass with P100 (I develop on V100)?

wat3rBro:
@slayton58 it happens during the first iteration. Could you verify if it works for e2e_mask_rcnn_X_101_32x8d_FPN_1x.yaml? On P100 I have R-50-C4/FPN running successfully without accuracy loss.

slayton58:
@wat3rBro I was having an issue with my branch due to some of the pre-trained model URLs having changed ([this is the relevant commit](https://github.com/facebookresearch/maskrcnn-benchmark/commit/49b96394c14f554965de08c37240d4580753b424)) -- I rebased (and pushed) against latest master and I can start training on a single V100 now.

wat3rBro:
@slayton58 Does it run properly on your machine at least for the first few iterations?

slayton58:
Yes. Are you still having issues?
```
2019-01-10 14:11:12,900 maskrcnn_benchmark.trainer INFO: Start training
2019-01-10 14:11:25,448 maskrcnn_benchmark.trainer INFO: eta: 1 day, 7:21:44  iter: 20  loss: 2.5158 (3.3709)  loss_classifier: 0.5149 (1.5699)  loss_box_reg: 0.0209 (0.0326)  loss_mask: 0.8690 (1.0426)  loss_objectness: 0.6652 (0.6536)  loss_rpn_box_reg: 0.0419 (0.0723)  time: 0.4033 (0.6273)  data: 0.0040 (0.1115)  lr: 0.000448  max mem: 6004
INFO:maskrcnn_benchmark.trainer:eta: 1 day, 7:21:44  iter: 20  loss: 2.5158 (3.3709)  loss_classifier: 0.5149 (1.5699)  loss_box_reg: 0.0209 (0.0326)  loss_mask: 0.8690 (1.0426)  loss_objectness: 0.6652 (0.6536)  loss_rpn_box_reg: 0.0419 (0.0723)  time: 0.4033 (0.6273)  data: 0.0040 (0.1115)  lr: 0.000448  max mem: 6004
2019-01-10 14:11:33,937 maskrcnn_benchmark.trainer INFO: eta: 1 day, 2:17:19  iter: 40  loss: 1.5658 (2.6424)  loss_classifier: 0.2581 (1.0473)  loss_box_reg: 0.0278 (0.0597)  loss_mask: 0.7494 (0.9107)  loss_objectness: 0.3690 (0.5331)  loss_rpn_box_reg: 0.0280 (0.0915)  time: 0.4112 (0.5259)  data: 0.0046 (0.0584)  lr: 0.000482  max mem: 6124
INFO:maskrcnn_benchmark.trainer:eta: 1 day, 2:17:19  iter: 40  loss: 1.5658 (2.6424)  loss_classifier: 0.2581 (1.0473)  loss_box_reg: 0.0278 (0.0597)  loss_mask: 0.7494 (0.9107)  loss_objectness: 0.3690 (0.5331)  loss_rpn_box_reg: 0.0280 (0.0915)  time: 0.4112 (0.5259)  data: 0.0046 (0.0584)  lr: 0.000482  max mem: 6124
```

fmassa:
One thing to keep in mind: @wat3rBro what version of CUDNN are you using that gives the problem?

wat3rBro:
@fmassa I'm still on CuDNN 7.1.2 because of our infra.

slayton58:
@wat3rBro Any progress on your side? If we want to try and track down more details on exactly where the failure's happening on your end you could try running with `CUDA_LAUNCH_BLOCKING=1` and see where the error is reported (if you're not running with that env variable then pytorch's normal async error checking is happening and the error location can be _really_ misleading)

wat3rBro:
@slayton58 it does work with 7.4.2. `Floating point exception (core dumped)` was what I got on 7.1.4. Are you going to fix the old version?

slayton58:
@wat3rBro could you run with the following please and give me the resulting cudnn.log file please -- I can use it to file an internal bug (or email me at slayton (at) nvidia (dot) com if it's too big)

```
export CUDA_LAUNCH_BLOCKING=1
export CUDNN_LOGINFO_DBG=1
export CUDNN_LOGDEST_DBG=./cudnn.log
<normal single GPU run command here>
```

slayton58:
@wat3rBro Any progress?

wat3rBro:
Hey @slayton58, I've sent you the log, sorry if I didn't notify you.

slayton58:
@wat3rBro Hmm, I haven't seen anything come through my email - what address did you send it from?

wat3rBro:
Maybe it's blocked, I just sent you again from my personal email.

ClimbsRocks:
Curious, what's the current state of this PR? Some of the X-152 models I'm training with are huge and slow. I've seen notable improvements from training with NVIDIA's half-precision code and PyTorch in the past.

zimenglan-sysu-512:
hi @slayton58 
curious, what's the current state of this PR?

LeviViana:
I've made some tests on 4 x 2080Ti. I didn't experience any improvement in training time. In my tests, I set `_C.DYTPE = "float16"` vs `_C.DYTPE = "float32"`. I tested :

- `e2e_faster_rcnn_R_50_FPN_1x.yaml` for 500 iterations
- `e2e_mask_rcnn_X_101_32x8d_FPN_1x.yaml` for 500 iterations

@slayton58 what could be going wrong on these tests ? Could you provide some benchmarks on the improvements you've achieved ?

slayton58:
@zimenglan-sysu-512 I'm still waiting on a merge. The PR should still be up-to-date and compatible with recent refactors of `apex.amp`.

@LeviViana Speedup is dependent on network and batch size per GPU -- `e2e_mask_rcnn_R_50_FPN_1x.yaml` is the config most tested with per-GPU batch sizes of 1, 2, 4. Based on the numbers from the model zoo and my iterations pasted above, the speedup on this config is ~0.45/0.4 = 12.5%. ResNeXt backbones (the second config you mention) don't seem to benefit this second from fp16, and I've never actually run faster-RCNN so I can't comment there.

LeviViana:
Thanks for your quick answer. Indeed, my precedent tests were based on a 1 img/gpu setup. I re-ran some other tests on the `e2e_faster_rcnn_R_50_FPN_1x.yaml` architecture, and I got about:
- 10% speedup with 2 imgs/gpu
- 20% speedup with 3 imgs/gpu

zimenglan-sysu-512:
thanks @slayton58 
hi @fmassa do u still have a plan to merge the PR?

fmassa:
Sorry for the delay in merging, and thanks @slayton58 !

zimenglan-sysu-512:
hi @slayton58 
i have a question, will `apex` change the type of `labels` in `targets`, e.g. change `torch.int64` to `torch.float32`?

slayton58:
@zimenglan-sysu-512 No, labels will stay as `torch.int64`

bendidi:
I don't know if it was taken into account but `apex` throws an error when trying to run inference on a cpu only machine

what i installed: 

```Docker
RUN conda install pytorch-cpu=1.1.0 torchvision-cpu=0.3.0 -c pytorch

# install apex
RUN git clone https://github.com/NVIDIA/apex.git /apex
RUN cd /apex && git checkout 14e34f7f89967dcbe5876b8bf416e311dd90b9dd && python setup.py install --cpp_ext

# install PyTorch maskrcnn-benchmark
RUN git clone https://github.com/Sterblue/maskrcnn-benchmark.git /maskrcnn-benchmark \
    && cd /maskrcnn-benchmark \
    && git checkout 5c41f1225208c3cd22b9c4734fa1d89f3f4de592 \
    && python setup.py build develop
```

error: 

```
    if utils.get_cuda_version() >= (9, 1, 0):
  File "/miniconda/envs/py36/lib/python3.6/site-packages/apex-0.1-py3.6-linux-x86_64.egg/apex/amp/utils.py", line 9, in get_cuda_version
    return tuple(int(x) for x in torch.version.cuda.split('.'))
AttributeError: 'NoneType' object has no attribute 'split'
```

PR #202, opened 2018-11-23
Dockerfile with jupyter notebook support

keineahnung2345:
Hello, 
In this PR, I add support to jupyter notebook, so that one can program from remote.
To make this work, one needs to run the image with `nvidia-docker run -td -p 8888:8888 -e PASSWORD=<password> maskrcnn-benchmark-jupyter`, in which one sets the password which is used by jupyter notebook. And then one can go to localhost:8888 and start programming.

I put the 3 repos(vision, cocoapi, maskrcnn-benchmark) inside the newly added folder "/notebooks", tell me if it needs to be changed.


fmassa:
cc @miguelvr as I don't know much about Dockerfiles

miguelvr:
I never used jupyter notebooks inside a docker container tbh... is there really a need for this?

keineahnung2345:
@miguelvr I agree it's optional. But with it, one can use his own browser to program and don't need to go into the docker and then use vim to do the same thing.

fmassa:
I'm ok with merging this, but as I mentioned before I'm not super familiar with Docker so it would be good if someone else could review this.

miguelvr:
> @miguelvr I agree it's optional. But with it, one can use his own browser to program and don't need to go into the docker and then use vim to do the same thing.

you don't have necessarily to go into the docker and use vim... you can mount directories, use solutions like pycharm, etc.

But sure, it is a nice option to have.

I just have one question: why is the password thing necessary?

fmassa:
@miguelvr can approve the PR once you are good with it? I'll merge once you approve

docker/docker-jupyter/jupyter_notebook_config.py
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ==============================================================================
+import os
+from IPython.lib import passwd
+
+#c = c  # pylint:disable=undefined-variable
+c = get_config()
```
miguelvr: where is `get_config()` coming from?

docker/docker-jupyter/jupyter_notebook_config.py
```diff
@@ -0,0 +1,32 @@
+# Copyright 2015 The TensorFlow Authors. All Rights Reserved.
```
miguelvr: hmmm, is this okay @fmassa ?

fmassa: I think?
cc @soumith

docker/docker-jupyter/jupyter_notebook_config.py
keineahnung2345: In [Overview of the IPython configuration system](https://ipython.org/ipython-doc/3/development/config.html), it says:
> While in a configuration file, to get a reference to this object, simply call the get_config() function, which is available in the global namespace of the script.

keineahnung2345:
@fmassa Could you please merge this PR now?

fmassa:
Thanks!

PR #205, opened 2018-11-23
fix opencv installation in dockerfile

miguelvr:
@keineahnung2345 your approach was indeed better, the conda installation of opencv is not compiled against all of opencv's functionalities.

This PR fixes the opencv installation


fmassa:
Thanks @miguelvr !

PR #211, opened 2018-11-26
Add script files for training and testing

xuw080:


fmassa:
Hi,

Thanks for the PR.

I think that each user might have a specific way of launching jobs for training.
For example, the user might want to pass extra arguments via `$@`, or pass a specific directory for saving the experiments.

Because of that, I think it's easier to let it up to the user to write their own scripts for dispatching to the underlying code, specially given that the bash commands are already given in the README.

I'm closing the PR because of what I've just mentioned, but let me know if you disagree.

PR #224, opened 2018-11-28
Fix indexing error in nms kernel

ppwwyyxx:
Here it is indexing a cuda tensor with CPU indices. It used to work, but 
after https://github.com/pytorch/pytorch/commit/006505bb8f9dcf0f38b32518308d071c8a1ccec6 it results in memory corruption.


fmassa:
Thanks for the PR!
Don't we need to pay attention to the device in which we will be shipping it to cuda?

ppwwyyxx:
It can run on multiple GPUs now, but that's probably only because there is only one GPU per process?
I probably should use the cuda device of other tensors.

fmassa:
I'll test this patch tomorrow and will merge it afterwards

PR #207, opened 2018-11-23
add support for pascal voc dataset and evaluate

fmassa:
Copy of #131 with a few fixes.

I didn't manage to make things work properly on that PR, I'm very sorry for that @lufficc ,
Your commits can be seen in the history.

Issue #168, opened 2018-11-18
Train on cityscapes


jinfagang:
How to train on cityscapes dataset? Does any one trained on it?

fmassa:
Training on a new dataset can be done by following the instructions in https://github.com/facebookresearch/maskrcnn-benchmark/issues/15 , and by using the coco-like dataset format for cityscapes, which can be found in https://github.com/facebookresearch/Detectron/blob/master/tools/convert_cityscapes_to_coco.py
Alternatively, you can write your own dataset for that.

Also, in https://github.com/facebookresearch/maskrcnn-benchmark/issues/166 @ltnghia managed to train on cityscapes.
Maybe he can share the procedure with us, and maybe (even better) send a PR adding support for CityScapes?

ltnghia:
I cannot click on "Create pull request".
@jinfagang: You can use my code in #166, it's correct. Follow the procedure, you definitely can train on Cityscapes.

fmassa:
@ltnghia this is weird. Did you push the code to any public branch of your fork of mask RCNN benchmark?

PR #232, opened 2018-11-29
Support more datasets

henrywang1:
I would like to make a proposal for the issues about supporting other datasets. e.g. #168 and #207.
#168 doesn't have a clear solution so far, and #207 doesn't support evaluating segmentation results.

Inspired by [Detectron](https://github.com/facebookresearch/Detectron/blob/ffb87f4cf73276ef53feb65195e6a262d8ee5bae/detectron/core/config.py#L280), I propose that adding a new config option, "DATASETS.FORCE_USE_JSON_ANNOTATION".
If the option is True, maskrcnn-benchmark will use COCO API to load the specified dataset and evaluate the result with the COCO styled AP. (i.e. AP, AP50, AP75, APs, APm, APl for bbox and segm)

In addition to the newly added option, I also
- Provide the steps to run on other datasets (maskrcnn_benchmark/data/Readme.md)
- Add the example configuration for Cityscapes
- Organize the codes for converting Cityscapes Datasets to COCO style
  - Reference: [[1]](https://github.com/facebookresearch/Detectron/blob/97241609a88e50e51d5966cc4d8868bcbff628cf/tools/convert_cityscapes_to_coco.py), [[2]](https://github.com/facebookresearch/Detectron/issues/111)

You could try the below command options to evaluate the changes.
```
python train_net.py --config-file "configs/pascal_voc/e2e_mask_rcnn_R_50_FPN_1x.yaml" DATASETS.FORCE_USE_JSON_ANNOTATION True
python train_net.py --config-file "configs/e2e_faster_rcnn_R_50_FPN_1x.yaml" DATASETS.FORCE_USE_JSON_ANNOTATION True
python train_net.py --config-file "configs/e2e_mask_rcnn_R_50_FPN_1x.yaml" DATASETS.FORCE_USE_JSON_ANNOTATION True
```

Thanks !


maskrcnn_benchmark/data/datasets/voc.py
```diff
@@ -41,7 +41,7 @@ class PascalVOCDataset(torch.utils.data.Dataset):
     )
 
     def __init__(self, data_dir, split, use_difficult=False, transforms=None):
-        self.root = data_dir
+        self.root = data_dir[:-10]  # strip "JPEGImages"
```
fmassa: This looks very fragile

maskrcnn_benchmark/engine/trainer.py
```diff
@@ -101,8 +101,10 @@ def do_train(
                     memory=torch.cuda.max_memory_allocated() / 1024.0 / 1024.0,
                 )
             )
-        if iteration % checkpoint_period == 0 or iteration == max_iter:
+        if iteration % checkpoint_period == 0:
```
fmassa: Is this change intended?

maskrcnn_benchmark/config/defaults.py
```diff
@@ -272,3 +276,4 @@
 _C.OUTPUT_DIR = "."
 
 _C.PATHS_CATALOG = os.path.join(os.path.dirname(__file__), "paths_catalog.py")
+
```
fmassa: Can you remove this newline?

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -21,18 +20,38 @@ class DatasetCatalog(object):
...
             "coco/annotations/instances_valminusminival2014.json",
         ),
-        "voc_2007_trainval": ("voc/VOC2007", 'trainval'),
-        "voc_2007_test": ("voc/VOC2007", 'test'),
-        "voc_2012_train": ("voc/VOC2012", 'train'),
-        "voc_2012_trainval": ("voc/VOC2012", 'trainval'),
-        "voc_2012_val": ("voc/VOC2012", 'val'),
-        "voc_2012_test": ("voc/VOC2012", 'test'),
-
+        "voc_2007_train": (
```
fmassa: instead of replacing all the uses that we had before with json-style, what do you think about having a different name tag for COCO-style dataset, like `voc_2007_trainval_cocostyle`?

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -21,18 +20,38 @@ class DatasetCatalog(object):
...
+            "cityscapes/annotations/instancesonly_filtered_gtFine_val.json"),
+        "cityscapes_fine_instanceonly_seg_test": (
+            "cityscapes/images",
+            "cityscapes/annotations/instancesonly_filtered_gtFine_test.json")
     }
 
     @staticmethod
     def get(name):
-        if "coco" in name:
+        if "coco" in name or cfg.DATASETS.FORCE_USE_JSON_ANNOTATION:
```
fmassa: I'd rather not have this option, and instead have the name tag be representative of what the dataset is or how it should be read.

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -48,7 +67,7 @@ def get(name):
             attrs = DatasetCatalog.DATASETS[name]
             args = dict(
                 data_dir=os.path.join(data_dir, attrs[0]),
-                split=attrs[1],
+                split=name[9:],
```
fmassa: This looks very fragile

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -2,11 +2,10 @@
 """Centralized catalog of paths."""
 
 import os
-
+from . import cfg
```
fmassa: If we don't have the `FORCE_USE_JSON_ANNOTATION`, this import won't be necessary

tools/instances2dict_with_polygons.py
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/python
```
fmassa: Can we add those two files in a subfolder inside `tools/`?

henrywang1: updated

maskrcnn_benchmark/config/defaults.py
henrywang1: updated

maskrcnn_benchmark/engine/trainer.py
henrywang1: There was a merge error before, and it is fixed now.

henrywang1:
Hi @fmassa 
I agree with all your comments, so I made some changes.
While I was modifying the code, I think that

Instead of having two entries in `path_catalog.py`
```python
"voc_2012_train": ("voc/VOC2012", 'train'),
"voc_2012_train_cocostyle": ("voc/VOC2012/JPEGImages", "voc/VOC2012/Annotations/pascal_test2012.json"),
```
If we use dictionary to replace tuple, it will be more readable
```python
"voc_2012_train": {
    "data_dir": "voc/VOC2012",
    "img_dir": "voc/VOC2012/JPEGImages",
    "ann_file": "voc/VOC2012/Annotations/pascal_train2012.json",
    "split": "train"
},
```
In addition, I think using code to append *_cocostyle entries to DATASETS is clearer.
```python
coco_style_datasets = {
    key + "_cocostyle": value for key, value in DATASETS.items() if "coco" not in key
}	    }
DATASETS={**DATASETS, **coco_style_datasets}
```

What do you think about all the updates?
Btw, I will modify `maskrcnn_benchmark/data/Readme.md`, once we reach an agreement.

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -6,38 +6,86 @@
...
+        "cityscapes_fine_instanceonly_seg_test": {
+            "img_dir": "cityscapes/images",
+            "ann_file": "cityscapes/annotations/instancesonly_filtered_gtFine_test.json"
+        }
+    }
 
+    coco_style_datasets = {
+        key + "_cocostyle": value for key, value in DATASETS.items() if "coco" not in key
     }
+    DATASETS={**DATASETS, **coco_style_datasets}
```
fmassa: This is not Py2-compatible.
Can you do instead
```python
DATASETS.update(coco_style_datasets)
```

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -6,38 +6,86 @@
...
 
     @staticmethod
     def get(name):
         if "coco" in name:
             data_dir = DatasetCatalog.DATA_DIR
             attrs = DatasetCatalog.DATASETS[name]
             args = dict(
-                root=os.path.join(data_dir, attrs[0]),
-                ann_file=os.path.join(data_dir, attrs[1]),
+                root=os.path.join(data_dir, attrs["img_dir"]),
```
fmassa: much better, thanks!

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -6,38 +6,86 @@
 
 class DatasetCatalog(object):
```
fmassa: in the top of this file and `defaults.py` there is a `100644 → 100755`. I believe this is a change in permission (via chmod). Can you verify and change back to what it was?

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -6,38 +6,86 @@
...
+            "ann_file": "cityscapes/annotations/instancesonly_filtered_gtFine_val.json"
+        },
+        "cityscapes_fine_instanceonly_seg_test": {
+            "img_dir": "cityscapes/images",
+            "ann_file": "cityscapes/annotations/instancesonly_filtered_gtFine_test.json"
+        }
+    }
 
+    coco_style_datasets = {
+        key + "_cocostyle": value for key, value in DATASETS.items() if "coco" not in key
```
fmassa: what about also adding as constraints that there is a `img_dir` and an `ann_file` as fields of those elements?
Because I can imagine that people will want to add new datasets that are not going to be using the coco evaluation API, and this will force it use this API right away.
---

Actually, on a second thought. I think that this hides away from the user what are the datasets that support COCO-style loading and evaluation. We also have repeated entries for cityscapes, which basically do the same thing.
Let's maybe keep things a bit more verbose, but at least clearer on the intent?

So, what about:
- add `voc_xxx` and `voc_xxx_cocostyle`
- add only `cityscapes_xxx_cocostyle` (given that we don't support native loaders / evaluators for now)

Are you ok with that? I think that the amount of code-duplication will be fairly small, given that we will only have a handful of datasets that will be manually implemented.

### Longer term
Have a 100% dataset-agnostic evaluation tool, so that those details can be hidden away from the user.

maskrcnn_benchmark/config/paths_catalog.py
henrywang1: I've changed back. Thanks for the careful review!

maskrcnn_benchmark/config/paths_catalog.py
henrywang1: That sounds a good idea. 
I believe things will be more verbose and simple by doing that, and users could understand the meanings easier.

If it do have too many datasets in the future, or we've built the dataset-agonistic evaluation tool,
we could think about the new representation by then.

maskrcnn_benchmark/config/paths_catalog.py
```diff
@@ -6,28 +6,80 @@
 
 class DatasetCatalog(object):
     DATA_DIR = "datasets"
-
```
fmassa: nit: `config/defaults.py` still have the change of permissions, can you change that?

henrywang1:
Thanks for the review and all the comments.
I've updated the code, the format and logic has changed to:
- voc_xxx and voc_xxx_cocostyle
- cityscapes_xxx_cocostyle

PR #246, opened 2018-12-04
fix tqdm usage

CoinCheung:
Maybe this is the behavior that the authors would like to see, I guess.

### DIFF
--- a/maskrcnn_benchmark/engine/inference.py
+++ b/maskrcnn_benchmark/engine/inference.py
@@ -17,7 +17,7 @@ def compute_on_dataset(model, data_loader, device):
     model.eval()
     results_dict = {}
     cpu_device = torch.device("cpu")
-    for i, batch in tqdm(enumerate(data_loader)):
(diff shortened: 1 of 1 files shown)


fmassa:
Is there a difference?

CoinCheung:
I think so. On my platform of archlinux with python3.7, the following two pieces of code behaves differently (one has process bar, while the other does not).
```
from tqdm import tqdm
s = 1
for i, n in enumerate(tqdm(range(100000))):
    s += i
```
```
from tqdm import tqdm
s = 1
for i, n in tqdm(enumerate(range(100000))):
    s += i
```

fmassa:
Oh, I see the difference now, thanks.

While this change would be ok, the fact that we use `tqdm` here is purely because I didn't bother changing it to a more logging-friendly approach back then.

While having the progress bar would be nice, it would also clutter much more some logs.
Anyway, I think we can merge this and remove the usage of tqdm later on.

PR #271, opened 2018-12-14
edited CUDA files to avoid Windows 10 build errors

Maosef:
As per issue 254: https://github.com/facebookresearch/maskrcnn-benchmark/issues/254

Added functions to replace THCeilDiv function in ROIAlign and ROIPool, successfully builds on Windows 10.


fmassa:
Fixed in a simpler way via https://github.com/facebookresearch/maskrcnn-benchmark/pull/409

Issue #15, opened 2018-10-25
How to finetune from pretrained detectron models with different number of classes?


wangg12:
## ❓ Questions and Help
Is there a config option to load pretrained coco models for finetuning? The last layers where  the number of classes may be different, so those weights should not be loaded.

fmassa:
Hi,

There currently isn't an off-the-shelf option in the config for that.
I see two easy options:
1 - from a python interpreter, load the pre-trained files that you want to use, and delete from the `state_dict` the keys corresponding to the last layer. The exact naming depends on the model architecture, but for boxes the name will [end with a `cls_score` and `bbox_pred`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/modeling/roi_heads/box_head/roi_box_predictors.py#L16-L17), and for masks it will [end with `mask_fcn_logits`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/modeling/roi_heads/mask_head/roi_mask_predictors.py#L24).
2 - Clone the code-base and modify the names of the two variables that I pointed out to be something else, like `cls_score_mine` etc. This will work out of the box, and you can modify the NUM_CLASSES in the config without clashes.

I think we could provide a functionality to perform 1 for the users, given a cfg file and a path to a model weight. That could be a possible improvement on top of what we currently have.

What do you think?

fmassa:
Yeah, option 1 is definitely simpler for the user (even if there are only a few lines to change here and there ;-) )

I'll prepare a PR adding support for this functionality, but I'm not 100% sure of what the API should look like, nor the best fix for it.

## API
Should we have a function that acts on the weights and creates a new set of weights file? Or should we add an extra config argument, to make it a single step function? If we add an argument (which seems simpler for the user), would it be ambiguous?

## Implementation

For the possible fixes, we could hard-code the possible names for the layers that shouldn't be loaded (as I mentioned before). But this is not super robust if the user changes their module names (which they can, if they want).

Another possible implementation is to not load the weights for the entire `predictor`. This is effectively the most robust way, as the predictor was designed to be only the "last layer".
[...]

PR #286, opened 2018-12-18
add finetune from detectron guide

wangg12:
#15

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,7 +152,7 @@ from maskrcnn_benchmark.structures.bounding_box import BoxList
 class MyDataset(object):
     def __init__(self, ...):
         # as you would do normally
-    
+
     def __getitem__(self, idx):
         # load the image as a PIL Image
         image = ...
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -192,6 +192,12 @@ While the aforementioned example should work for training, we leverage the
 cocoApi for computing the accuracies during testing. Thus, test datasets
 should currently follow the cocoApi for now.
 
+## Finetuning from Detectron weights on custom datasets
+Create a script `tools/trim_detectron_model.py` like [here](https://gist.github.com/wangg12/aea194aa6ab6a4de088f14ee193fd968).
+You can decide which keys to be removed and which keys to be kept by modifying the script.
+
+Then you can simply point the converted model path in the config file by changing `MODEL.WEIGHT`.
```
fmassa: Can you also point to the relevant issue?
Something like

> For further information, please refer to https://github.com/facebookresearch/maskrcnn-benchmark/issues/15

wangg12: Sure.

README.md
```diff
@@ -192,6 +192,14 @@ While the aforementioned example should work for training, we leverage the
...
+
+For further information, please refer to #15.
```
fmassa: I think you need to point to the link, something like

```
[#15](https://github.com/facebookresearch/maskrcnn-benchmark/issues/15)
```

wangg12: OK

fmassa:
@DIYer22 I think that the error as we have now is the safest option at the moment.
We already log a number of things during training, and a warning could easily be missed, but training would go on but results would be worse.

PR #299, opened 2018-12-21
Update checkpoint.py

DIYer22:
if edit "{save_dir}/last_checkpoint" by vim, "\n" will auto append to the end of file. which will raise 

```
FileNotFoundError: [Errno 2] No such file or directory: 'save_dir/model_0060000.pth\n'
```

### DIFF
--- a/maskrcnn_benchmark/utils/checkpoint.py
+++ b/maskrcnn_benchmark/utils/checkpoint.py
@@ -79,6 +79,7 @@ class Checkpointer(object):
         try:
             with open(save_file, "r") as f:
                 last_saved = f.read()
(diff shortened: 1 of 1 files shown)


maskrcnn_benchmark/utils/checkpoint.py
```diff
@@ -79,6 +79,7 @@ def get_checkpoint_file(self):
         try:
             with open(save_file, "r") as f:
                 last_saved = f.read()
+                last_saved = last_saved.replace("\n", "")
```
fmassa: what about using `.strip()` instead, would that be more robust (for example, on Windows)?

PR #300, opened 2018-12-21
Update test_net.py, checkpoint.py to support "--load-pth"

DIYer22:
Split `Checkpointer.load` to support choice which .pth file to load for test

### DIFF
--- a/maskrcnn_benchmark/utils/checkpoint.py
+++ b/maskrcnn_benchmark/utils/checkpoint.py
@@ -57,6 +57,9 @@ class Checkpointer(object):
             # no checkpoint could be found
             self.logger.info("No checkpoint found. Initializing model from scratch")
             return {}
+        return self.load_pth_file(f)
+
+    def load_pth_file(self, f):
         self.logger.info("Loading checkpoint from {}".format(f))
         checkpoint = self._load_file(f)
         self._load_model(checkpoint)
--- a/tools/test_net.py
+++ b/tools/test_net.py
@@ -26,6 +26,13 @@ def main():
         metavar="FILE",
         help="path to config file",
     )
+    parser.add_argument(
+        "--load-pth",
+        default="",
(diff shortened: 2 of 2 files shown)


fmassa:
Thanks for the PR!
Could you just expand a bit on why this is needed?

DIYer22:
According to [this line](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/utils/checkpoint.py#L54)

If os.path.isfile("OUTPUT_DIR/last_checkpoint"), MODEL.WEIGHT will be override by "OUTPUT_DIR/last_checkpoint"

When user what to test another .pth file(MODEL.WEIGHT), the loaded weights will be  "OUTPUT_DIR/last_checkpoint", rather than MODEL.WEIGHT

fmassa:
Yes, but if you change `OUTPUT_DIR` to point to a different path (which you might want to do to avoid overwriting data in the current training folder), then the problem disappears, right?

DIYer22:
😂, It's seens  that no document to suggest user to change `OUTPUT_DIR` when test. 
I think there is some users have the same problem like I got, because those users may put a experment(include train and test) into one `OUTPUT_DIR`  in commonsense.

fmassa:
I meant that if you test on the same folder as when training, it will create a new subfolder with the results of the experiments using the latest checkpoint.
But if you just re-run it again with your patch, I believe it will overwrite the results from this subfolder, as it will have the same name for the files, right?

It might be a good idea to mention a possible solution in the README or something, what do you think?

DIYer22:
I think so, it's better to mention user that to change OUTPUT_DIR when want to test on  specific .pth file(MODEL.WEIGHT).

fmassa:
@DIYer22 would you be willing to send a PR improving the README? Thanks!

PR #344, opened 2019-01-15
use F.batch_norm in inference of FrozenBatchNorm2d 

wat3rBro:
this makes FrozenBatchNorm2d ONNX compatible

### DIFF
--- a/maskrcnn_benchmark/modeling/backbone/resnet.py
+++ b/maskrcnn_benchmark/modeling/backbone/resnet.py
@@ -244,6 +244,10 @@ class Bottleneck(nn.Module):
                 ),
                 norm_func(out_channels),
             )
+            for modules in [self.downsample,]:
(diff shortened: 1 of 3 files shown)


xxradon:
@wat3rBro How do you use export maskrcnn to onnx formate?

wat3rBro:
@xxradon it requires special handle of customized layers, one solution is adding another codepath which calls ONNX/Caffe2 ops directly.

wat3rBro:
move to https://github.com/facebookresearch/maskrcnn-benchmark/pull/384

zimenglan-sysu-512:
hi @wat3rBro 
can u share how to use onnx to export maskrcnn?
thanks

wat3rBro:
@zimenglan-sysu-512 I exported different ConvNets separately and assemble them together with specialized ops.

Issue #110, opened 2018-11-05
Support Group Normalization?


yaoliUoA:
## ❓ Questions and Help
Hi, 

Does this implementation supports group normailization? I think Detectron supports group normalization (https://github.com/facebookresearch/Detectron/blob/master/projects/GN/README.md). If not, any plans to add group normailzation in this package?

fmassa:
Hi,

We don't have Group Normalization implemented yet, but it would be good to have.
I'm currently working on other parts of the library, but contributions are more than welcome! :-)

zimenglan-sysu-512:
hi @fmassa 
do u have a plan to add GN? 
those [Detectron.pytorch](https://github.com/roytseng-tw/Detectron.pytorch), [Detectron-Cascade-RCNN](https://github.com/zhaoweicai/Detectron-Cascade-RCNN), and [PANet](https://github.com/ShuLiu1993/PANet), use python to implement it, while [Detectron](https://github.com/facebookresearch/Detectron) use caffe2. 
can u give me some ideas how to add GN to this benchmark?

fmassa:
It should be fairly straightforward, like changing `FrozenBatchNorm2d` by `GroupNorm`, plus a few architectural changes. Basically, [change this file](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/modeling/backbone/resnet.py), adding the new blocks in the registration in https://github.com/facebookresearch/maskrcnn-benchmark/blob/83849b4ed116ee30a4a3d94519f7ee29d8755b2d/maskrcnn_benchmark/modeling/backbone/resnet.py#L294-L305

After that, it would also require loading a pre-trained imagenet model trained using GN.

Let me know if you need more information

zimenglan-sysu-512:
thanks @fmassa 
i follow what u said to add GN, and right now i try to run some experiments.
thanks again

fmassa:
Thanks, let me know how it goes, and if it works, could you send a PR adding this functionality?

zimenglan-sysu-512:
hi @fmassa 
i add `GN` from [PANet](https://github.com/ShuLiu1993/PANet), and retrain [retinanet](https://github.com/facebookresearch/maskrcnn-benchmark/pull/102/) using R50, only boost 0.5%  mAP (from 36.4 to 36.9). and training my own dataset using GN drops a little. 
maybe some bugs in the code i add with GN.

fmassa:
[...]

PR #346, opened 2019-01-15
add GroupNorm

zimenglan-sysu-512:
when u want to train from scratch, please be cafeful about OOM


zimenglan-sysu-512:
#110

jackshaw:
Relu in the conv3x3 module returned by function make_conv3x3 may cause redundancy. There is another a relu function in MaskRCNNFPNFeatureExtractor.forward.

maskrcnn_benchmark/modeling/backbone/fpn.py
```diff
@@ -11,7 +12,10 @@ class FPN(nn.Module):
     order, and must be consecutive
     """
 
-    def __init__(self, in_channels_list, out_channels, top_blocks=None):
+    def __init__(
+        self, in_channels_list, out_channels, 
+        top_blocks=None, use_gn=False
```
fmassa: What about, instead of passing a `use_gn` argument, pass a `conv_block` argument?
Then, you could keep the code almost as is:
```python
inner_block_module = conv_block(in_channels, out_channels, 1)
layer_block_module = conv_block(out_channels, out_channels, 3, 1, 1)
```

and you can create functions that return what you want:
```python
def conv_with_kaiming_uniform(...):
    module = nn.Conv2d(...)
    nn.init.kaiming....
    return module
```
and the same for the `group_norm`?

maskrcnn_benchmark/modeling/backbone/resnet.py
```diff
@@ -289,13 +298,125 @@ def forward(self, x):
         x = F.relu_(x)
         x = F.max_pool2d(x, kernel_size=3, stride=2, padding=1)
         return x
+        return x
```
fmassa: this seems duplicate?

maskrcnn_benchmark/modeling/roi_heads/mask_head/roi_mask_feature_extractors.py
```diff
@@ -32,17 +34,17 @@ def __init__(self, cfg):
...
         next_feature = input_size
         self.blocks = []
         for layer_idx, layer_features in enumerate(layers, 1):
             layer_name = "mask_fcn{}".format(layer_idx)
-            module = Conv2d(next_feature, layer_features, 3, stride=1, padding=1)
-            # Caffe2 implementation uses MSRAFill, which in fact
-            # corresponds to kaiming_normal_ in PyTorch
-            nn.init.kaiming_normal_(module.weight, mode="fan_out", nonlinearity="relu")
-            nn.init.constant_(module.bias, 0)
+            module = make_conv3x3(next_feature, layer_features,
```
fmassa: This will currently break loading of previously saved models, because you always return a `nn.Sequential` here.
Could you instead return a `nn.Conv2d` if it's not `gn`?
Also, as mentioned, there is a reduntant ReLU in `make_conv3x3`.

maskrcnn_benchmark/modeling/backbone/resnet.py
```diff
@@ -289,13 +298,125 @@ def forward(self, x):
...
+        if self.downsample is not None:
+            identity = self.downsample(x)
+
+        out += identity
+        out = F.relu_(out)
+
+        return out
+
+
+class StemWithGN(nn.Module):
```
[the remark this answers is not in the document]
zimenglan-sysu-512: i agree with u, i change it like this:
```python

class Bottleneck(nn.Module):
    def __init__(
        self,
        in_channels,
        bottleneck_channels,
        out_channels,
        num_groups,
        stride_in_1x1,
        stride,
        norm_func
    ):
        super(BottleneckWithFixedBatchNorm, self).__init__()

        self.downsample = None
        if in_channels != out_channels:
            self.downsample = nn.Sequential(
                Conv2d(
                    in_channels, out_channels, kernel_size=1, stride=stride, bias=False
                ),
                norm_func(out_channels),
            )

        # The original MSRA ResNet models have stride in the first 1x1 conv
        # The subsequent fb.torch.resnet and Caffe2 ResNe[X]t implementations have
        # stride in the 3x3 conv
        stride_1x1, stride_3x3 = (stride, 1) if stride_in_1x1 else (1, stride)

        self.conv1 = Conv2d(
            in_channels,
            bottleneck_channels,
            kernel_size=1,
            stride=stride_1x1,
            bias=False,
        )
        self.bn1 = norm_func(bottleneck_channels)
        # TODO: specify init for the above

        self.conv2 = Conv2d(
            bottleneck_channels,
            bottleneck_channels,
            kernel_size=3,
            stride=stride_3x3,
            padding=1,
            bias=False,
            groups=num_groups,
        )
        self.bn2 = norm_func(bottleneck_channels)

        self.conv3 = Conv2d(
            bottleneck_channels, out_channels, kernel_size=1, bias=False
        )
        self.bn3 = norm_func(out_channels)

    def forward(self, x):
        identity = x

        out = self.conv1(x)
        out = self.bn1(out)
        out = F.relu_(out)

        out = self.conv2(out)
        out = self.bn2(out)
        out = F.relu_(out)

        out0 = self.conv3(out)
        out = self.bn3(out0)

        if self.downsample is not None:
            identity = self.downsample(x)

        out += identity
        out = F.relu_(out)

        return out


class BaseStem(nn.Module):
    def __init__(self, cfg, norm_func):
        super(StemWithFixedBatchNorm, self).__init__()

        out_channels = cfg.MODEL.RESNETS.STEM_OUT_CHANNELS

        self.conv1 = Conv2d(
            3, out_channels, kernel_size=7, stride=2, padding=3, bias=False
        )
        self.bn1 = norm_func(out_channels)

    def forward(self, x):
        x = self.conv1(x)
        x = self.bn1(x)
        x = F.relu_(x)
        x = F.max_pool2d(x, kernel_size=3, stride=2, padding=1)
        return x


class BottleneckWithFixedBatchNorm(Bottleneck):
    def __init__(
        self,
        in_channels,
        bottleneck_channels,
        out_channels,
        num_groups=1,
        stride_in_1x1=True,
        stride=1,
    ):
        super(BottleneckWithFixedBatchNorm, self).__init__(
            in_channels=in_channels,
            bottleneck_channels=bottleneck_channels,
            out_channels=out_channels,
            num_groups=num_groups,
            stride_in_1x1=stride_in_1x1,
            norm_func=FrozenBatchNorm2d
        )


class StemWithFixedBatchNorm(BaseStem):
    def __init__(self, cfg):
        super(StemWithFixedBatchNorm, self).__init__(
            cfg, norm_func=FrozenBatchNorm2d
        )


class BottleneckWithGN(Bottleneck):
    def __init__(
        self,
        in_channels,
        bottleneck_channels,
        out_channels,
        num_groups=1,
        stride_in_1x1=True,
        stride=1,
        dilation=1
    ):
        super(BottleneckWithGN, self).__init__(
            in_channels=in_channels,
            bottleneck_channels=bottleneck_channels,
            out_channels=out_channels,
            num_groups=num_groups,
            stride_in_1x1=stride_in_1x1,
            norm_func=group_norm
        )


class StemWithGN(BaseStem):
    def __init__(self, cfg):
        super(StemWithGN, self).__init__(cfg, norm_func=group_norm)
```

maskrcnn_benchmark/utils/c2_model_loading.py
```diff
@@ -47,6 +47,18 @@ def _rename_basic_resnet_weights(layer_keys):
     layer_keys = [k.replace(".branch1.", ".downsample.0.") for k in layer_keys]
     layer_keys = [k.replace(".branch1_bn.", ".downsample.1.") for k in layer_keys]
 
+    # GroupNorm
+    layer_keys = [k.replace("conv1.gn.s", "gn1.weight") for k in layer_keys]
```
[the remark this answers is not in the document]
zimenglan-sysu-512: got u

maskrcnn_benchmark/layers/group_norm.py
```diff
@@ -0,0 +1,92 @@
...
+Group Normalization Layer from PANet
+url: https://github.com/ShuLiu1993/PANet
+"""
+
+import torch
+import torch.nn as nn
+from maskrcnn_benchmark.config import cfg
+
+
+class GroupNorm(nn.Module):
```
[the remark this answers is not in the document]
zimenglan-sysu-512: ok

maskrcnn_benchmark/layers/group_norm.py
```diff
@@ -0,0 +1,92 @@
+"""
+Group Normalization Layer from PANet
+url: https://github.com/ShuLiu1993/PANet
+"""
+
+import torch
+import torch.nn as nn
+from maskrcnn_benchmark.config import cfg
```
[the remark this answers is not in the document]
zimenglan-sysu-512: yes, i remove `layers/group_norm.py`, and move `gn` construction function into `modeling/make_layers.py`, like this:
```python

def get_group_gn(dim, dim_per_gp, num_groups):
    """get number of groups used by GroupNorm, based on number of channels."""
    assert dim_per_gp == -1 or num_groups == -1, \
        "GroupNorm: can only specify G or C/G."

    if dim_per_gp > 0:
        assert dim % dim_per_gp == 0, \
            "dim: {}, dim_per_gp: {}".format(dim, dim_per_gp)
        group_gn = dim // dim_per_gp
    else:
        assert dim % num_groups == 0, \
            "dim: {}, num_groups: {}".format(dim, num_groups)
        group_gn = num_groups

    return group_gn


def _group_norm():
    def gn_layer_from_cfg(out_channels, affine=True, divisor=1):
        out_channels = out_channels // divisor
        dim_per_gp = cfg.GROUP_NORM.DIM_PER_GP // divisor
        num_groups = cfg.GROUP_NORM.NUM_GROUPS // divisor
        eps = cfg.GROUP_NORM.EPSILON # default: 1e-5
        return torch.nn.GroupNorm(
            get_group_gn(out_channels, dim_per_gp, num_groups), 
            out_channels, 
            eps, 
            affine
        )
    return gn_layer_from_cfg
group_norm = _group_norm()
```

maskrcnn_benchmark/layers/group_norm.py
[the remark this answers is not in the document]
jackshaw: I agree with your idea. I found that this line of code`x.view(N,G,-1)` in function [group_norm](https://github.com/facebookresearch/maskrcnn-benchmark/pull/346/files/4877e36b9b1241f98890ac141e8e4400f25dbea4#diff-9f1d682529e708ec42b0d7c72cad2353) sometimes may cause exception. During test,maybe there is no bounding box outputed from detection branch for a single image. So the shape of outputed bounding boxes is (0,n). Furthermore, if we send this (0,n) tensor or array to mask branch with GN, expectation will happen when running to  `x.view(N,G,-1)`.

maskrcnn_benchmark/modeling/backbone/resnet.py
```diff
@@ -289,13 +298,125 @@ def forward(self, x):
...
+        super(StemWithGN, self).__init__()
+
+        out_channels = cfg.MODEL.RESNETS.STEM_OUT_CHANNELS
+
+        self.conv1 = Conv2d(
+            3, out_channels, kernel_size=7, stride=2, padding=3, bias=False
+        )
+        self.gn1 = gn_layer_from_cfg(out_channels)
+
+        # for l in [self.conv1,]:
```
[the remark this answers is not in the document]
zimenglan-sysu-512: i will remove it

maskrcnn_benchmark/modeling/backbone/resnet.py
zimenglan-sysu-512: it is duplicated, i will remove it

maskrcnn_benchmark/modeling/roi_heads/mask_head/roi_mask_feature_extractors.py
zimenglan-sysu-512: understand. i will keep it as before

maskrcnn_benchmark/modeling/backbone/fpn.py
zimenglan-sysu-512: i agree with u, it can simply the code

zimenglan-sysu-512:
hi @fmassa 
i follow what u say to change the code and test it using the trained model from `model_zoo`. for both `configs/e2e_mask_rcnn_R_50_FPN_1x.yaml`  and `configs/e2e_faster_rcnn_R_50_FPN_1x.yaml`, they match the results. 
in the next week, i will run the experiments for `GN`, please wait for the results.
thanks.

maskrcnn_benchmark/modeling/backbone/resnet.py
```diff
@@ -178,6 +199,7 @@ def _make_stage(
     num_groups,
     stride_in_1x1,
     first_stride,
+    dilation=1
```
fmassa: just double-checking if dilation is 1 here

maskrcnn_benchmark/modeling/backbone/resnet.py
```diff
@@ -230,24 +260,25 @@ def __init__(
...
+        self.bn1 = norm_func(bottleneck_channels)
         # TODO: specify init for the above
 
         self.conv2 = Conv2d(
             bottleneck_channels,
             bottleneck_channels,
             kernel_size=3,
             stride=stride_3x3,
-            padding=1,
+            padding=dilation, # dilation * (kernel_size - 1) // 2,
```
fmassa: just checking if this comment should be removed or not?

maskrcnn_benchmark/modeling/make_layers.py
```diff
@@ -0,0 +1,134 @@
...
+        num_groups = cfg.MODEL.GROUP_NORM.NUM_GROUPS // divisor
+        eps = cfg.MODEL.GROUP_NORM.EPSILON # default: 1e-5
+        return torch.nn.GroupNorm(
+            get_group_gn(out_channels, dim_per_gp, num_groups), 
+            out_channels, 
+            eps, 
+            affine
+        )
+    return gn_layer_from_cfg
+group_norm = _group_norm()
```
fmassa: we can probably remove this extra indirection, as we don't use `_group_norm` anywhere else?
By that, I mean making `group_norm` be simply `gn_layer_from_cfg`

maskrcnn_benchmark/modeling/make_layers.py
```diff
@@ -0,0 +1,134 @@
...
+
+
+def make_conv3x3(
+    in_channels, 
+    out_channels, 
+    dilation=1, 
+    stride=1, 
+    use_gn=False,
+    use_relu=False,
```
fmassa: looks like this option is not used?

maskrcnn_benchmark/modeling/make_layers.py
```diff
@@ -0,0 +1,134 @@
...
+        return nn.Sequential(fc, group_norm(hidden_dim))
+    fc = nn.Linear(dim_in, hidden_dim)
+    nn.init.kaiming_uniform_(fc.weight, a=1)
+    nn.init.constant_(fc.bias, 0)
+    return fc
+
+
+def conv_with_kaiming_uniform(use_gn=False):
+    def make_conv(
+        in_channels, out_channels, kernel_size, stride=1, dilation=1, use_relu=False
```
fmassa: I think the `use_relu` option is not used? Do we want to keep it around?

maskrcnn_benchmark/modeling/make_layers.py
```diff
@@ -0,0 +1,134 @@
...
+    return group_gn
+
+
+def _group_norm():
+    def gn_layer_from_cfg(out_channels, affine=True, divisor=1):
+        out_channels = out_channels // divisor
+        dim_per_gp = cfg.MODEL.GROUP_NORM.DIM_PER_GP // divisor
+        num_groups = cfg.MODEL.GROUP_NORM.NUM_GROUPS // divisor
+        eps = cfg.MODEL.GROUP_NORM.EPSILON # default: 1e-5
+        return torch.nn.GroupNorm(
```
fmassa: Question: should `GroupNorm` be initialized on a specific way? For example, by setting the bias to zero and the weights to 1?

maskrcnn_benchmark/modeling/backbone/resnet.py
zimenglan-sysu-512: right, here dilation is 1

maskrcnn_benchmark/modeling/backbone/resnet.py
zimenglan-sysu-512: it can be removed, but can be a explanation for using dilation param. i will remove it.

maskrcnn_benchmark/modeling/make_layers.py
zimenglan-sysu-512: sometimes it is convenient to put these ops `conv`, `bn` or `gn` and  `relu`  into `nn.Sequential`, i would like to change it like this: 
```python
def conv_with_kaiming_uniform(use_gn=False, use_relu=False):
    def make_conv(
        in_channels, out_channels, kernel_size, stride=1, dilation=1
    ):
        module = Conv2d(
            in_channels, 
            out_channels, 
            kernel_size=kernel_size, 
            stride=stride, 
            padding=dilation * (kernel_size - 1) // 2, 
            dilation=dilation, 
            bias=False if use_gn else True
        )
        # Caffe2 implementation uses XavierFill, which in fact
        # corresponds to kaiming_uniform_ in PyTorch
        nn.init.kaiming_uniform_(module.weight, a=1)
        if not use_gn:
            nn.init.constant_(module.bias, 0)
        if use_gn:
            if use_relu:
                return nn.Sequential(
                    module, group_norm(out_channels), nn.ReLU(inplace=True)
                )
            else:
                return nn.Sequential(
                    module, group_norm(out_channels)
                )
        if use_relu:
            return nn.Sequential(module, nn.ReLU(inplace=True))
        else:
            return module
    return make_conv
```
what do u think?

maskrcnn_benchmark/modeling/make_layers.py
zimenglan-sysu-512: it seems GroupNorm has been [initialized](https://pytorch.org/docs/stable/_modules/torch/nn/modules/normalization.html#GroupNorm).

maskrcnn_benchmark/modeling/backbone/resnet.py
fmassa: oh ok, it was not clear to me, it looked like you were trying to use one or the other

maskrcnn_benchmark/modeling/backbone/resnet.py
fmassa: Ok, I was just wondering if Detectron used 1 for dilation here in both cases, [and seems that they do use it](https://github.com/facebookresearch/Detectron/blob/7c0ad88fc0d33cf0f698a3554ee842262d27babf/detectron/modeling/ResNet.py#L293-L302), so it's good, thanks!

maskrcnn_benchmark/modeling/make_layers.py
fmassa: Ok, I just double checked in [Detectron and they initialize it to bias zero and weights 1](https://github.com/facebookresearch/Detectron/blob/7c0ad88fc0d33cf0f698a3554ee842262d27babf/detectron/modeling/detector.py#L447-L450), so this is good, thanks

maskrcnn_benchmark/modeling/make_layers.py
fmassa: Sounds good, we can keep this

maskrcnn_benchmark/modeling/make_layers.py
fmassa: Yes, this makes more sense to me to let the `use_relu` be in the `conv_with_kaiming_uniform`

configs/gn_baselines/e2e_faster_rcnn_R_50_FPN_1x_gn.yaml
```diff
@@ -0,0 +1,48 @@
...
+    POOLER_SAMPLING_RATIO: 2
+    FEATURE_EXTRACTOR: "FPN2MLPFeatureExtractor"
+    PREDICTOR: "FPNPredictor"
+DATASETS:
+  TRAIN: ("coco_2014_train", "coco_2014_valminusminival")
+  TEST: ("coco_2014_minival",)
+DATALOADER:
+  SIZE_DIVISIBILITY: 32
+SOLVER:
+  # Assume 4 gpus
```
fmassa: Can we make it consistent and assume that all scripts are for 8 GPUs?

I saw that some of the configs you added are for 4, others for 8, it can be confusing

maskrcnn_benchmark/modeling/make_layers.py
```diff
@@ -0,0 +1,131 @@
...
+    if kaiming_init:
+        nn.init.kaiming_normal_(
+            conv.weight, mode="fan_out", nonlinearity="relu"
+        )
+    else:
+        torch.nn.init.normal_(conv.weight, std=0.01)
+    if not use_gn:
+        nn.init.constant_(conv.bias, 0)
+    if use_gn:
+        if use_relu:
```
fmassa: What do you think about the following to reduce the amount of indentation and code duplication:
```python
module = [conv]
if use_gn:
    module.append(group_norm(out_channels)
if use_relu:
    module.append(nn.ReLU(inplace=True)
if len(module) > 1:
    return nn.Sequential(*module)
return conv
```
This simplifies a bit the codepath, and is simpler in my opinion

maskrcnn_benchmark/modeling/make_layers.py
```diff
@@ -0,0 +1,131 @@
...
+            padding=dilation * (kernel_size - 1) // 2, 
+            dilation=dilation, 
+            bias=False if use_gn else True
+        )
+        # Caffe2 implementation uses XavierFill, which in fact
+        # corresponds to kaiming_uniform_ in PyTorch
+        nn.init.kaiming_uniform_(module.weight, a=1)
+        if not use_gn:
+            nn.init.constant_(module.bias, 0)
+        if use_gn:
```
fmassa: same comment here, I think this can be simplified a bit.

configs/gn_baselines/e2e_faster_rcnn_R_50_FPN_1x_gn.yaml
zimenglan-sysu-512: ok

maskrcnn_benchmark/modeling/make_layers.py
zimenglan-sysu-512: got u

zimenglan-sysu-512:
hi @fmassa 
i run `configs/gn_baselines/scratch_e2e_mask_rcnn_R_50_FPN_Xconv1fc_3x_gn.yaml`, the results are as below:
```
bbox
Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.388
Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.593
Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.424
Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.236
Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.410
Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.505
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.320
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.507
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.531
Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.347
Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.563
Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.673

mask
Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.350
Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.564
Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.373
Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.170
Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.370
Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.515
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.299
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.463
Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.482
Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.286
Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.517
Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.637
```
a little bit performance drops, when compared with [Rethinking ImageNet Pre-training](https://arxiv.org/pdf/1811.08883.pdf)
```
for bbox:  38.8% v.s. 39.5%
for mask: 35.0% v.s. 35.2%
```

what do u think?

fmassa:
I think this looks great!

The difference might be due to the fact that we don't properly initialize the ResNet backbone as they did, but I think this is good enough!

Thanks a lot!

ppwwyyxx:
Just saw this from the commits. 
Another difference from the papers (in addition to the initialization issue above): in GN paper we never applied GN after any FC layers. We did not use it in the 2MLP head and we did not use it after the FC layer in the 4conv1fc head, though I expect they behave similar in accuracy. This issue can probably be spotted if the code can be tested against the GN models in Detectron model zoo.
Also, if the configs are not yet able to reproduce the results from Detectron, it's best to make a note in the config, to avoid disappointing other users who would naturally expect them to behave the same.

fmassa:
Those are great points @ppwwyyxx
@zimenglan-sysu-512 do you mind addressing the above points?

zimenglan-sysu-512:
hi @ppwwyyxx 
thanks for pointing it. can u provide the link to what u say, because i can't find it. and also i will disable `gn` in the fc layer of the 4conv1fc head explictly.

hi @fmassa 
i am willing to do it, but what shoud l do? add a `README` in the `configs/gn_baselines/` to point out? or something else i can do?

fmassa:
@zimenglan-sysu-512 here is what @ppwwyyxx was mentioning:
you added GN in https://github.com/facebookresearch/maskrcnn-benchmark/pull/346/files#diff-553631744ef7c9e38523e4dbda64ac37R134 , while the detectron implementation didn't, see https://github.com/facebookresearch/Detectron/blob/7c0ad88fc0d33cf0f698a3554ee842262d27babf/detectron/modeling/fast_rcnn_heads.py#L175-L177
This is also present in the https://github.com/facebookresearch/maskrcnn-benchmark/pull/346/files#diff-553631744ef7c9e38523e4dbda64ac37R68 (and probably elsewhere).

I think the best here would be to send a follow-up PR that removes those GN layers, and also more carefully check the ResNet initializations (I didn't bother adding the right ones because they were always overwritten by pre-trained weights).

@zimenglan-sysu-512 Would you be able to address those points?

Thanks!

fmassa:
@zimenglan-sysu-512 both sounds good to me.

For the initialization, I'd follow Detectron initialization, which can be found in https://github.com/facebookresearch/Detectron/blob/master/detectron/modeling/ResNet.py

Caffe2 seems to use by [default `XavierFill` and zero bias](https://github.com/pytorch/pytorch/blob/master/caffe2/python/helpers/conv.py#L75), which corresponds to `kaiming_uniform_` in PyTorch https://github.com/facebookresearch/maskrcnn-benchmark/blob/af132c51cac15db975b06bec03601d7ec1cb6410/maskrcnn_benchmark/modeling/make_layers.py#L108-L109

zimenglan-sysu-512:
hi @fmassa i re-train the `configs/gn_baselines/scratch_e2e_mask_rcnn_R_50_FPN_Xconv1fc_3x_gn.yaml` and get `38.8&35.1`, which is still a little gap compared with `detectron`.

fmassa:
Hi @zimenglan-sysu-512,

At this point I don't know where the difference might come from, I'd need to have a closer look.

If you don't have any ideas on where the difference might come from, then it might be good to add a note in the configs mentioning the accuracies that you obtain, so that people are aware of this difference?

zimenglan-sysu-512:
hi @fmassa 
i agree with u, if i have no ideas where the difference might come from, it is good to add a note in the configs to mention the performance. 
the only difference i find right now, detectron use `STRIDE_1X1=False`, where i set `STRIDE_IN_1X1` to be True.

fmassa:
Oh, that's a problem! Because the pre-trained models from ResNeXt were trained with PyTorch style, this is something that should be fixed I think.

zimenglan-sysu-512:
hi @fmassa 
u can see [here](https://github.com/facebookresearch/Detectron/blob/master/configs/04_2018_gn_baselines/scratch_e2e_mask_rcnn_R-50-FPN_3x_gn.yaml#L21). so i set  `STRIDE_IN_1X1` to be `False` in the yaml file?

fmassa:
Yes, you should set it to be false, as we use the same pre-trained models for classification as Detectron.

zimenglan-sysu-512:
ok 
i will add it later and re-run the trainings.

fmassa:
Can you already send a PR, and update it when the trainings are over?

zimenglan-sysu-512:
ok, after trainings, i will send a PR.

zimenglan-sysu-512:
hi @fmassa 
i run `e2e_mask_rcnn_R_50_FPN_Xconv1fc_2x_gn.yaml`, and get `box=40.2 (v.s 40.3)` and `mask=36.0 (v.s 35.7)`. and when `scratch_e2e_mask_rcnn_R_50_FPN_Xconv1fc_3x_gn.yaml` is finished, i will report the results here.
i guess the main little gap is caused by `STRIDE_IN_1X1`, which need to be set `False`, when `backbone` use `GN`.

fmassa:
@zimenglan-sysu-512 awesome! Can you send a PR changing the `STRIDE_IN_1x1`? Looks like this was indeed the cause for most of the difference!

zimenglan-sysu-512:
hi @fmassa 
i wait for the results of `scratch_e2e_mask_rcnn_R_50_FPN_Xconv1fc_3x_gn`. after that, i will send a PR.

fmassa:
I would say that it would be better to directly send a PR, given that this is definitely a difference between the implementations and I'm positive the sooner we fix this the better it will be for the community trying to use those models

zimenglan-sysu-512:
hi @fmassa 
sorry for the delay.
the result of `scratch_e2e_mask_rcnn_R_50_FPN_Xconv1fc_3x_gn` is 38.9 for bbox and 35.1 for mask, which is still a little gap compared with `detectron` (39.5 for bbox and 35.2 for mask).

fmassa:
@zimenglan-sysu-512 thanks a lot for the heads up!

In this case, I'd recommend adding a warning in the beginning of the config files that do not reproduce the results, so that users are aware of this difference. Can you send a PR?

zimenglan-sysu-512:
hi @fmassa 
i find that from Detectron, [this line](https://github.com/facebookresearch/Detectron/blob/8c1e2afb74b3b5efaa8dc198b1bac81398593634/detectron/core/config.py#L620) sets weight_decay to be 0.0, and [this line](https://github.com/facebookresearch/Detectron/blob/8c1e2afb74/detectron/modeling/optimizer.py#L104) uses it to init the params of gn. and in the yaml files, they do not set it, which means that they use the default value 0.0.
maybe hte reason of the results not matching the Detectron is that we should deal with `weight_decay` carefully for GN layer.

zimenglan-sysu-512:
hi @fmassa 
is it possible that the performance gap is caused by the `weight_decay` for gn layers. in Detectron, it sets the weight_decay 0.0 for gn in [config](https://github.com/facebookresearch/Detectron/blob/8c1e2afb74b3b5efaa8dc198b1bac81398593634/detectron/core/config.py#L620) and [optimizer](https://github.com/facebookresearch/Detectron/blob/8c1e2afb74/detectron/modeling/optimizer.py#L104).

fmassa:
@zimenglan-sysu-512 maybe? I'm not sure though

kevinj22:
I'm not sure how the caffe2 group norm models work (i.e if they track running batch stats, and if they do over all parallel processes), but I believe the pytorch group norm does NOT track these stats. In the event I am incorrect and it does track stats, pytorch also does not track synchronized stats over many GPU's. If the caffe2 model does this, it's running means and variances will be a better estimation for validation time.

PR #336, opened 2019-01-12
fixup cv2.findContours to only return 2 values

aaronlelevier:
This PR is to fix an error that I ran in to when running [demo/Mask_R-CNN_demo.ipynb](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/demo/Mask_R-CNN_demo.ipynb)

The PR removes the first `_` assigned variable. Based on the docs for opencv `cv2.findContours` only returns 2 values. Here's the docs link:

[https://docs.opencv.org/master/d3/dc0/group__imgproc__shape.html#gadf1ad6a0b82947fa1fe3c3d497f260e0](https://docs.opencv.org/master/d3/dc0/group__imgproc__shape.html#gadf1ad6a0b82947fa1fe3c3d497f260e0
)

Here is the stack trace that I got from the current code on *master*:

```
ValueError                                Traceback (most recent call last)
<ipython-input-10-ba06eae97bcc> in <module>
      1 # compute predictions
----> 2 predictions = coco_demo.run_on_opencv_image(image)
      3 imshow(predictions)

~/github/maskrcnn-benchmark/demo/predictor.py in run_on_opencv_image(self, image)
    177         result = self.overlay_boxes(result, top_predictions)
    178         if self.cfg.MODEL.MASK_ON:
--> 179             result = self.overlay_mask(result, top_predictions)
    180         result = self.overlay_class_names(result, top_predictions)
    181 

~/github/maskrcnn-benchmark/demo/predictor.py in overlay_mask(self, image, predictions)
    289             thresh = mask[0, :, :, None]
    290             _, contours, hierarchy = cv2.findContours(
--> 291                 thresh, cv2.RETR_TREE, cv2.CHAIN_APPROX_SIMPLE
    292             )
    293             image = cv2.drawContours(image, contours, -1, color, 3)

ValueError: not enough values to unpack (expected 3, got 2)
```

I was using these package versions when I ran into the error:

```
Python 3.6.7
opencv-python==4.0.0.21
```

PR #354, opened 2019-01-19
Add helper function "cv2_utils.findContours" to maintain version compatibility between versions 3 and 4

aaronlelevier:
## Summary

Small PR to fix the version compatibility problem with `cv2.findContours` returning either 2 or 3 values depending on the cv2 version.

I replaced all occurrences of `cv2.findContours` with the helper.

I put the code in `utils/` dir. Seemed like the best place for it, but please advise.

## Testing

I tested locally by installing my branch locally using:

```
pip install -U git+https://github.com/aaronlelevier/maskrcnn-benchmark.git@train
```

Then ran the `demo/Mask_R-CNN_demo.ipynb ` and it worked as expected, even though my `cv2` version is `opencv-python==4.0.0.21`, so this is now compatible with either version.

## References

fixes #336 

#339 also referenced this error but the issue is already closed

### DIFF
--- a/demo/predictor.py
+++ b/demo/predictor.py
@@ -8,6 +8,7 @@ from maskrcnn_benchmark.utils.checkpoint import DetectronCheckpointer
 from maskrcnn_benchmark.structures.image_list import to_image_list
 from maskrcnn_benchmark.modeling.roi_heads.mask_head.inference import Masker
 from maskrcnn_benchmark import layers as L
+from maskrcnn_benchmark.utils import cv2_util
 
 
 class COCODemo(object):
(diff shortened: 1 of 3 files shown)


aaronlelevier:
Okay, I removed the other files / commits, so this PR is now only related to the `cv2.findCountours` change. 

I meant to change branches and was still working on this branch, and didn't realize that it was also my PR branch!  

All, should be good now. Please lmk if you have any other feedback? I did see the comment to support cv2 versions 1-4 in #336 's suggested snippet , but I thought maybe only versions 3-4 is relevant, so my code is for that. 

Thanks

Issue #135, opened 2018-11-09
[Tutorial] Conda + gcc5 on Ubuntu 14.04


jinyu121:
## ❓ Questions and Help

Well, I have a server with Ubuntu 14.04, whose `gcc` is 4.8.X. Hey, it is 2018 now, the latest version is 18.10, and we are using Ubuntu 14.04, without `sudo`.

After some tries, here is a short tutorial.

```
conda create --name maskrcnn-benchmark
conda activate maskrcnn-benchmark

conda install gcc-5 gcc-5-mp -c psi4
conda install ipython ninja cython matplotlib
conda install pytorch-nightly torchvision -c pytorch 
pip install yacs pycocotools 

git clone https://github.com/facebookresearch/maskrcnn-benchmark.git
cd maskrcnn-benchmark
```

Then, open `setup.py`, add `extra_link_args=['-L/usr/lib/x86_64-linux-gnu/']` in `ext_modules`>`extension`, so this section will like

```
    ext_modules = [
        extension(
            "maskrcnn_benchmark._C",
            sources,
            include_dirs=include_dirs,
            define_macros=define_macros,
            extra_compile_args=extra_compile_args,
            extra_link_args=['-L/usr/lib/x86_64-linux-gnu/']  # Yes~ I am here
        )
    ]
```

Finally, 

```
python setup.py build develop
```

and have fun.

fmassa:
Thanks for the information!

We have included in the `TROUBLESHOOTING` section of the README a part that mentions [how to install a newest version of gcc](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/TROUBLESHOOTING.md#segmentation-fault-core-dumped-when-running-the-library).

Let us know if we could improve the documentation with further information.

PR #368, opened 2019-01-22
Update setup.py

ZhangGongjie:
When using gcc/g++ from anaconda environment, we should add this into setup.py, so that we can successfully build the project.

If we do not add, the compilation will fail.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,6 +50,7 @@ def get_extensions():
             include_dirs=include_dirs,
             define_macros=define_macros,
             extra_compile_args=extra_compile_args,
+            extra_link_args=['-L/usr/lib/x86_64-linux-gnu/'],
         )
     ]
 


fmassa:
Are you sure about this change? Wouldn't it be enough to have `gcc` be linked to your particular gcc that you are using?

ZhangGongjie:
I tried to set up the project with anaconda environment. This works.

This is an Anaconda internal issue, see https://github.com/cocodataset/cocoapi/issues/94.

ZhangGongjie:
Plus, adding this line definitely would not hurt the compilation of the codes.

fmassa:
I believe this line won't work for many other systems either, like OSX or Windows, that's why I'm hesitant about this PR

This seems similar to https://github.com/facebookresearch/maskrcnn-benchmark/issues/135, which points to the TROUBLESHOOTING section https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/TROUBLESHOOTING.md#segmentation-fault-core-dumped-when-running-the-library

ZhangGongjie:
Oh yeah... You are right. This should specifically works for Ubuntu OS. Is there a way to detect users' OS?

fmassa:
There should be with the `os` package.

Can you tell me what was the error you were having without this line?

ZhangGongjie:
g++ -pthread -shared -B /export/home/zhanggongjie/anaconda3/compiler_compat -L/export/home/zhanggongjie/anaconda3/lib -Wl,-rpath=/export/home/zhanggongjie/anaconda3/lib -Wl,--no-as-needed -Wl,--sysroot=/ build/temp.linux-x86_64-3.6/export/home/zhanggongjie/github/maskrcnn-benchmark-adaptiveFPN10/maskrcnn_benchmark/csrc/vision.o build/temp.linux-x86_64-3.6/export/home/zhanggongjie/github/maskrcnn-benchmark-adaptiveFPN10/maskrcnn_benchmark/csrc/cpu/ROIAlign_cpu.o build/temp.linux-x86_64-3.6/export/home/zhanggongjie/github/maskrcnn-benchmark-adaptiveFPN10/maskrcnn_benchmark/csrc/cpu/nms_cpu.o build/temp.linux-x86_64-3.6/export/home/zhanggongjie/github/maskrcnn-benchmark-adaptiveFPN10/maskrcnn_benchmark/csrc/cuda/ROIAlign_cuda.o build/temp.linux-x86_64-3.6/export/home/zhanggongjie/github/maskrcnn-benchmark-adaptiveFPN10/maskrcnn_benchmark/csrc/cuda/nms.o build/temp.linux-x86_64-3.6/export/home/zhanggongjie/github/maskrcnn-benchmark-adaptiveFPN10/maskrcnn_benchmark/csrc/cuda/ROIPool_cuda.o -L/usr/local/cuda/lib64 -lcudart -o build/lib.linux-x86_64-3.6/maskrcnn_benchmark/_C.cpython-36m-x86_64-linux-gnu.so
/export/home/zhanggongjie/anaconda3/compiler_compat/ld: cannot find -lm
/export/home/zhanggongjie/anaconda3/compiler_compat/ld: cannot find -lpthread
/export/home/zhanggongjie/anaconda3/compiler_compat/ld: cannot find -lc
collect2: error: ld returned 1 exit status
error: command 'g++' failed with exit status 1

botcs:
Closing due to inactivity

PR #377, opened 2019-01-23
use 'kaiming_uniform' to initialize resnet, disable gn after fc layer

zimenglan-sysu-512:


zimenglan-sysu-512:
hi @fmassa 
recently i need to do my own project, if i have time and resource, i will retrain from scratch and report the results here.
thanks

fmassa:
@zimenglan-sysu-512 thanks, if I find the time I'll kick some trainings as well and let you know

PR #383, opened 2019-01-24
use all_gather to gather results from all gpus

wat3rBro:
from @ppwwyyxx

### DIFF
--- a/maskrcnn_benchmark/utils/comm.py
+++ b/maskrcnn_benchmark/utils/comm.py
@@ -1,52 +1,46 @@
-# Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved.
 """
 This file contains primitives for multi-gpu communication.
 This is useful when doing distributed training.
 """
 
-import os
 import pickle
-import tempfile
 import time
 
 import torch
+import torch.distributed as dist
 
 
 def get_world_size():
-    if not torch.distributed.is_available():
+    if not dist.is_available():
         return 1
-    if not torch.distributed.is_initialized():
+    if not dist.is_initialized():
         return 1
-    return torch.distributed.get_world_size()
+    return dist.get_world_size()
 
 
 def get_rank():
-    if not torch.distributed.is_available():
+    if not dist.is_available():
         return 0
-    if not torch.distributed.is_initialized():
+    if not dist.is_initialized():
         return 0
-    return torch.distributed.get_rank()
+    return dist.get_rank()
 
 
 def is_main_process():
-    if not torch.distributed.is_available():
-        return True
-    if not torch.distributed.is_initialized():
-        return True
-    return torch.distributed.get_rank() == 0
+    return get_rank() == 0
 
 
 def synchronize():
     """
-    Helper function to synchronize between multiple processes when
+    Helper function to synchronize (barrier) among all processes when
     using distributed training
     """
-    if not torch.distributed.is_available():
+    if not dist.is_available():
         return
-    if not torch.distributed.is_initialized():
+    if not dist.is_initialized():
         return
-    world_size = torch.distributed.get_world_size()
-    rank = torch.distributed.get_rank()
+    world_size = dist.get_world_size()
+    rank = dist.get_rank()
     if world_size == 1:
         return
 
@@ -55,7 +49,7 @@ def synchronize():
             tensor = torch.tensor(0, device="cuda")
         else:
             tensor = torch.tensor(1, device="cuda")
-        torch.distributed.broadcast(tensor, r)
+        dist.broadcast(tensor, r)
         while tensor.item() == 1:
             time.sleep(1)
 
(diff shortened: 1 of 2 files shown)


qianyizhang:
whats the advantage using one over the other?

maskrcnn_benchmark/utils/comm.py
```diff
@@ -64,94 +58,73 @@ def _send_and_wait(r):
...
+
+    data_list = []
+    for size, tensor in zip(size_list, tensor_list):
+        buffer = tensor.cpu().numpy().tobytes()[:size]
+        data_list.append(pickle.loads(buffer))
+
+    return data_list
+
+
+def reduce_dict(input_dict, average=True):
```
fmassa: nit: I believe this is not used anywhere (it probably is due to a refactor in the engine/trainer?)

maskrcnn_benchmark/utils/comm.py
```diff
@@ -64,94 +58,73 @@ def _send_and_wait(r):
...
+    max_size = max(size_list)
+
+    # receiving Tensor from all ranks
+    # we pad the tensor because torch all_gather does not support
+    # gathering tensors of different shapes
+    tensor_list = []
+    for _ in size_list:
+        tensor_list.append(torch.ByteTensor(size=(max_size,)).to("cuda"))
+    if local_size != max_size:
+        padding = torch.ByteTensor(size=(max_size - local_size,)).to("cuda")
```
fmassa: nit: this could generate NaN because the data is unnitialized. This per se doesn't affect the overall results because we remove the padded value, but I'm not sure if it could cause problems with `dist.all_gather`.

maskrcnn_benchmark/utils/comm.py
```diff
@@ -64,94 +58,73 @@ def _send_and_wait(r):
...
+    dist.all_gather(size_list, local_size)
+    size_list = [int(size.item()) for size in size_list]
+    max_size = max(size_list)
+
+    # receiving Tensor from all ranks
+    # we pad the tensor because torch all_gather does not support
+    # gathering tensors of different shapes
+    tensor_list = []
+    for _ in size_list:
+        tensor_list.append(torch.ByteTensor(size=(max_size,)).to("cuda"))
```
fmassa: nit: it would be good to use the new API for this:

```python
device=torch.device("cuda")
...
torch.empty((max_size,), dtype=torch.uint8, device=device
```

maskrcnn_benchmark/utils/comm.py
```diff
@@ -64,94 +58,73 @@ def _send_and_wait(r):
...
-        os.rmdir(tmp_dir)
-        return data_list
+
+    # serialized to a Tensor
+    buffer = pickle.dumps(data)
+    storage = torch.ByteStorage.from_buffer(buffer)
+    tensor = torch.ByteTensor(storage).to("cuda")
+
+    # obtain Tensor size of each rank
+    local_size = torch.IntTensor([tensor.numel()]).to("cuda")
```
fmassa: nit: It would be good to replace the usages of `IntTensor`/`ByteTensor` with their new constructs. In this case (because it's initialized):
```python
device = torch.device("cuda")
...
local_size = torch.tensor([tensor.numel()], dtype=torch.int32, device=device)
# or, if 0d tensors work with dist.all_gather
local_size = torch.tensor(tensor.numel(), dtype=torch.int32, device=device)
```

fmassa:
@qianyizhang the advantage is that it makes it possible to do multi-machine testing, which was not possible before.

qianyizhang:
@fmassa 
thanks. Also I have another somewhat related question
Is it possible to run multiple sessions of maskrcnn_benchmark (or even torch.distributed processes) on the same node?
for example i have a server with 8 gpus and i want to run 2 sets of experiments using 4 cards each.

I got the rendezvous complaining "RuntimeError: Address already in use", how to make it work?

fmassa:
@qianyizhang yes, it's possible, but you need to change the `master_addr` and the `master_port` in `torch.distributed.launch`, see https://github.com/pytorch/pytorch/blob/master/torch/distributed/launch.py#L164-L169

maskrcnn_benchmark/utils/comm.py
```diff
@@ -64,94 +58,73 @@ def _send_and_wait(r):
...
-            # cleanup
-            os.remove(file_path)
-        # cleanup
-        os.rmdir(tmp_dir)
-        return data_list
+
+    # serialized to a Tensor
+    buffer = pickle.dumps(data)
+    storage = torch.ByteStorage.from_buffer(buffer)
+    tensor = torch.ByteTensor(storage).to("cuda")
```
fmassa: Actually, @ppwwyyxx this will probably be problematic for large datasets, as we will run out of memory on the GPU when trying to perform this communication.

The idea that I had was to use shared memory on the CPU and communicate the address of the shared memory, but this doesn't work on the multiple-machine case.

ppwwyyxx: I agree. But is there any way to do all-gather on CPUs (given that the dist backend was initialized with "nccl")?

fmassa: With c10d, it is now possible to have more than one dist backend at a time. So one could potentially have one nccl backend and one mpi backend?

maskrcnn_benchmark/utils/comm.py
```diff
@@ -64,94 +58,73 @@ def _send_and_wait(r):
...
+
+    # obtain Tensor size of each rank
+    local_size = torch.IntTensor([tensor.numel()]).to("cuda")
+    size_list = [torch.IntTensor([0]).to("cuda") for _ in range(world_size)]
+    dist.all_gather(size_list, local_size)
+    size_list = [int(size.item()) for size in size_list]
+    max_size = max(size_list)
```
pietern: @wat3rBro You can use `dist.allreduce(local_size, op=dist.ReduceOp.MAX)` here for a little less code.

ppwwyyxx: The sizes of all ranks are needed later.

pietern: Ah, I see, because of pickle requiring the exact size and doesn't tolerate additional NULs. Thanks for clarifying.

maskrcnn_benchmark/utils/comm.py
yelantingfeng: Yes, I am using this communication codes in other task. Because of the large size of each `data`,  I get OOM error.

fmassa: @yelantingfeng I'd recommend either:
- reverting this change locally for now
- try creating a new process group with c10d which is on the CPU, and communicate this `data` on the CPU instead

liu09114: > Yes, I am using these communication codes in other task. Because of the large size of each `data`, I get OOM error.
I got the OOM error. Have you implement the second method recommended by @fmassa ?

JoyHuYY1412:
> This is much better, thanks a lot @wat3rBro and @ppwwyyxx !
> 
> I've left a few nits, but I'm merging this as this is already much better than before.

HI,  Can I use 'all_gather' to gather the weights of all gpus? I want to achieve in each batch, different gpu outputs different weights, and the loss will be calculated using all the weights.  When I use 'all_gather', I found the output accumulated weights loss the grad_fn.

maskrcnn_benchmark/utils/comm.py
yelantingfeng: > @yelantingfeng I'd recommend either:
> 
> * reverting this change locally for now
> * try creating a new process group with c10d which is on the CPU, and communicate this `data` on the CPU instead

I think we could just set a memory limit for this all_gather function. I implement this by splitting those ByteTensors into chunks. After some tests, I found my implementation can limit the total usage of memory to MiB level. Though not 100% precise limit, but I think it should be useful enough. I would be glad to send a PR if you think this is a good improvement for this repository.

PR #384, opened 2019-01-24
Frozen batch norm2d onnx support

wat3rBro:
add a separate path for exporting model via onnx


jordanxlj:
@wat3rBro, have you tried to export onnx model of maskrcnn faster rcnn? I have tried, but always failed for some <listcomp> isn't supported.

wat3rBro:
@jordanxlj  it's not onnx compatible yet at this moment.

PR #398, opened 2019-01-31
FIxed comment for # of RoIs per minibatch

isameer:
Since IMS_PER_BATCH is global, it shouldn't be multiplied with the # of GPUs.


txytju:
https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/config/defaults.py#L190 
I think this line should also be modified, right?

PR #402, opened 2019-02-01
Registry for RoI Box Predictors

Cysu:
- Add a registry `ROI_BOX_PREDICTOR`
- Use the registry in `roi_box_predictors.py`, replacing the local factory
- Minor changes in `structures/bounding_box.py`: when copying a box with fields, check if the field exists
- Minor changes in `logger.py`: make filename a optional argument with default value of `"log.txt"`

### DIFF
--- a/maskrcnn_benchmark/structures/bounding_box.py
+++ b/maskrcnn_benchmark/structures/bounding_box.py
@@ -232,7 +232,7 @@ class BoxList(object):
             area = box[:, 2] * box[:, 3]
         else:
             raise RuntimeError("Should not be here")
-            
+
         return area
 
     def copy_with_fields(self, fields):
@@ -240,7 +240,8 @@ class BoxList(object):
         if not isinstance(fields, (list, tuple)):
             fields = [fields]
         for field in fields:
-            bbox.add_field(field, self.get_field(field))
+            if self.has_field(field):
+                bbox.add_field(field, self.get_field(field))
         return bbox
 
     def __repr__(self):
--- a/maskrcnn_benchmark/structures/bounding_box.py
+++ b/maskrcnn_benchmark/structures/bounding_box.py
@@ -235,13 +235,15 @@ class BoxList(object):
 
         return area
 
-    def copy_with_fields(self, fields):
+    def copy_with_fields(self, fields, skip_missing=False):
         bbox = BoxList(self.bbox, self.size, self.mode)
         if not isinstance(fields, (list, tuple)):
             fields = [fields]
         for field in fields:
             if self.has_field(field):
                 bbox.add_field(field, self.get_field(field))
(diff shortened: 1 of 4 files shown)


maskrcnn_benchmark/structures/bounding_box.py
```diff
@@ -232,15 +232,16 @@ def area(self):
...
+
         return area
 
     def copy_with_fields(self, fields):
         bbox = BoxList(self.bbox, self.size, self.mode)
         if not isinstance(fields, (list, tuple)):
             fields = [fields]
         for field in fields:
-            bbox.add_field(field, self.get_field(field))
+            if self.has_field(field):
```
fmassa: Good point.

When we use `copy_with_fields` and a field is not present, should it silently ignore the field (causing an error down the line most probably?) or raise an explicit error here?

I'd lean towards raising an error here, thoughts?

Cysu: @fmassa Good catch! This check mainly aims at supporting images with missing labels (such as mask or keypoints etc.). If we raise an error here, the copy cannot be finished, and the users might need to implement another similar function by themselves.

That said, how about we add an extra argument `skip_missing=False` here? If True, it will ignore the fields if missing, otherwise by default it will raise an error.

fmassa: Does the rest of the code work if an image is missing a label? I'm curious, because I haven't taken that into account while writing the code.

If the rest of the code works fine by ignoring the label, then I'd be ok with a `skip_missing=False`. But then would this field be set to `True` in all occurrences of `copy_with_fields`?

Cysu: Currently, the method is used in computing the loss for [box_head](https://github.com/facebookresearch/maskrcnn-benchmark/blob/9cb251d340/maskrcnn_benchmark/modeling/roi_heads/box_head/loss.py#L43) and [mask_head](https://github.com/facebookresearch/maskrcnn-benchmark/blob/9cb251d340/maskrcnn_benchmark/modeling/roi_heads/mask_head/loss.py#L61). The rest of the code does not work with missing box or mask labels (if mask head is enabled). Therefore, having `skip_missing=False` by default will raise an error for these two heads. This will not break any current experiments on COCO.

When users develop new roi heads for new datasets, they can set `skip_missing=True` and handle the missing labels in the corresponding loss computation.

fmassa: Ok, this sounds good to me. Can you add the aforementioned argument, and then this will be good to go, thanks!

Cysu: Added the `skip_missing=False` argument in the revision. Thanks a lot for the discussion!

PR #411, opened 2019-02-06
Replace with get_img_info on coco_eval

rodrigoberriel:
I was writing a script and saw some of these TODOs on `coco_eval.py` and I agree that using `get_img_info(...)` should be preferred. As they were straightforward to implement, here we are. I kept the same "style" used on [voc_eval.py](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/data/datasets/evaluation/voc/voc_eval.py#L18-L22). I performed an evaluation before and after the changes and I had no problems, as expected.

Although the TODOs seemed to be related only with the `width` and `height`, if we could assume that `dataset` is an instance of `COCODataset`, we could also change `original_id` to `img_info["id"]`. I was not sure about this assumption, because there are `assert`s in the beginning of these functions that are commented out.

### DIFF
--- a/maskrcnn_benchmark/data/datasets/evaluation/coco/coco_eval.py
+++ b/maskrcnn_benchmark/data/datasets/evaluation/coco/coco_eval.py
@@ -72,9 +72,9 @@ def prepare_for_coco_detection(predictions, dataset):
         if len(prediction) == 0:
             continue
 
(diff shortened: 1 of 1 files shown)


fmassa:
Yes, we could replace `original_id` as well I believe

PR #429, opened 2019-02-13
Some cleaning and changing default argument from `boxlist_nms`

LeviViana:


maskrcnn_benchmark/csrc/cpu/ROIAlign_cpu.cpp
```diff
@@ -115,32 +115,28 @@ void ROIAlignForward_cpu_kernel(
...
     const int pooled_width,
     const int sampling_ratio,
     const T* bottom_rois,
-    //int roi_cols,
     T* top_data) {
-  //AT_ASSERT(roi_cols == 4 || roi_cols == 5);
-  int roi_cols = 5;
 
-  int n_rois = nthreads / channels / pooled_width / pooled_height;
+  AT_ASSERT(roi_cols == 5)
```
fmassa: given that we only support `roi_cols == 5`, is there really a point in exposing it to the user?

Issue #447, opened 2019-02-17
detectron dependency in tools


botcs:
## 🐛 Bug
As suggested in the [data/README.md](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/data/README.md) `tools/convert_cityscapes_to_coco.py` requires detectron to be installed.

## To Reproduce

1. Install `maskrcnn_benchmark` following the INSTALL.md
1. Install additional dependencies for the script: `pip install h5py scipy opencv-python`
1. run `python tools/cityscapes/convert_cityscapes_to_coco.py`

### Error:
`ModuleNotFoundError: No module named 'detectron'`

## Environment
PyTorch version: 1.0.0.dev20190207
Is debug build: No
CUDA used to build PyTorch: 10.0.130

OS: Ubuntu 16.04.5 LTS
GCC version: (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609
CMake version: version 3.5.1

Python version: 3.6
Is CUDA available: Yes
CUDA runtime version: Could not collect
GPU models and configuration: 
GPU 0: GeForce RTX 2080
GPU 1: GeForce RTX 2080

Nvidia driver version: 410.72
cuDNN version: Could not collect

Versions of relevant libraries:
[pip] numpy==1.16.1
[pip] torch==1.0.0.dev20190207
[pip] torchvision==0.2.1
[conda] blas                      1.0                         mkl  
[conda] mkl                       2019.1                      144  
[conda] mkl_fft                   1.0.10           py36ha843d7b_0  
[conda] mkl_random                1.0.2            py36hd81dba3_0  
[conda] pytorch-nightly           1.0.0.dev20190207 py3.6_cuda10.0.130_cudnn7.4.2_0    pytorch

fmassa:
Oh, you are totally right, thanks for spotting this!

The dependency on Detectron is really not necessary, and was probably an oversight from #232 

There are only a couple of places where detectron is needed, and only two functions are used, see https://github.com/facebookresearch/maskrcnn-benchmark/blob/e60f4ec8dc50531debcfd5ae671ea167b5b7a1d9/tools/cityscapes/convert_cityscapes_to_coco.py#L196-L198

We already have something similar to `bboxs_util.xyxy_to_xywh` in `BoxList` via the `convert` method. The second function can be taken from Detectron to avoid this dependency.

Contributions are more than welcome! :-)

[...]

PR #448, opened 2019-02-17
clarifying directory structure and fixing typo

botcs:
A few addition:
I added the top level directory `cityscapes` since the `tools/cityscapes/convert_cityscapes_to_coco.py` script has the directory structure `gtFine_trainvaltest/gtFine` hardcoded into it which is fine but was not clear at first.

Also added a **Note** to warn people to install detectron as well, since the script uses `detectron.utils.boxes` and `detectron.utils.segm` modules which has further dependencies in the detectron lib.

### DIFF
--- a/maskrcnn_benchmark/data/README.md
+++ b/maskrcnn_benchmark/data/README.md
@@ -70,19 +70,22 @@ ln -s /path/to/cityscapes datasets/data/cityscapes
 1. Download gtFine_trainvaltest.zip from https://www.cityscapes-dataset.com/downloads/ (login required)
 2. Extract it to /path/to/gtFine_trainvaltest
 ```
-gtFine_trainvaltest
-|_ gtFine
+cityscapes
+|_ gtFine_trainvaltest.zip
+|_ gtFine_trainvaltest
+   |_ gtFine
 ```
 3. Run the below commands to convert the annotations
+**NOTE: Detectron AND maskrcnn_benchmark is required**
 
 ```
 cd ~/github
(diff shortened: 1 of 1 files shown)


maskrcnn_benchmark/data/README.md
```diff
@@ -70,19 +70,22 @@ ln -s /path/to/cityscapes datasets/data/cityscapes
...
 ```
-gtFine_trainvaltest
-|_ gtFine
+cityscapes
+|_ gtFine_trainvaltest.zip
+|_ gtFine_trainvaltest
+   |_ gtFine
 ```
 3. Run the below commands to convert the annotations
+**NOTE: Detectron AND maskrcnn_benchmark is required**
```
fmassa: This sounds good to me, but from a glance at the implementation, it is really easy to get rid of the dependency on Detectron.

botcs: Lazily I have copied `boxes.py` and `segms.py` from detectron to the maskrcnn utils, but then I saw that in boxes there was another level of dependency - and I got like "meh... just quickly install detectron", but then I had to resolve the caffe2 stuff :D so yeah... what would you suggest?

I can trace all the functions back in the detectron lib and copy them together to the maskrcnn lib or follow along [#447 ](https://github.com/facebookresearch/maskrcnn-benchmark/issues/447#issuecomment-464658449)

fmassa: I think you should follow what I've mentioned in #447. There is really only one function that needs to be ported from Detectron, the other is already in this repo. Let me know if you have further questions!

PR #463, opened 2019-02-19
Supported FBNet architecture.

newstzpz:
* Added FBNet architecture.
  * FBNet is a general architecture definition to support [efficient architecture search](https://arxiv.org/abs/1812.03443) and MaskRCNN2GO.
  * Included various efficient building blocks (inverted residual, shuffle, shift, separate dw conv, dw upsampling etc.)
  * Supported building backbone, rpn, detection, keypoint and mask heads using efficient building blocks.
  * Architecture could be defined in `fbnet_modeldef.py` or in `cfg.MODEL.FBNET.ARCH_DEF` directly.
  * A few baseline architectures are included (MobileNet V2 etc.).
* Supported any feature map size for average pool in bbox head. 
* Used registry to register keypoint and mask heads.
* Passing in/out channels between modules when creating the model.
  * This simplifies the code to compute the input channels for feature extractors and makes the predictors independent of the backbone architectures.
  * Passed in_channels to rpn and head builders.
  * Set out_channels to model modules including backbone and feature extractors.
  * Moved cfg.MODEL.BACKBONE.OUT_CHANNELS to cfg.MODEL.RESNETS.BACKBONE_OUT_CHANNELS as it is not used by all architectures. Updated config files accordingly.
  * For new architecture modules, the return module needs to contain a field called 'out_channels' to indicate the output channel size.
* Added various unit tests.
  * build and run backbones.
  * build and run feature extractors.
  * build and run predictors.
* Added an unit test to verify all config files are loadable.
* Added unit test for box_coder and nms.


maskrcnn_benchmark/modeling/backbone/fbnet_builder.py
```diff
@@ -0,0 +1,822 @@
...
+            BatchNorm2d(C_in),
+            nn.ReLU(inplace=True),
+            Conv2d(C_in, C_out, 3, 1, 1, bias=False),
+            BatchNorm2d(C_out),
+        ]
+        super(CascadeConv3x3, self).__init__(*ops)
+        self.res_connect = (stride == 1) and (C_in == C_out)
+
+    def forward(self, x):
+        y = super().forward(x)
```
fmassa: nit: could you add `super(CascadeConv3x3, self).forward(x)` here, for better Py2 compatibility?

maskrcnn_benchmark/modeling/backbone/fbnet_builder.py
```diff
@@ -0,0 +1,822 @@
...
+            # shift
+            Shift(C_mid, 5, stride, 2),
+            # pw-linear
+            Conv2d(C_mid, C_out, 1, 1, 0, bias=False),
+            BatchNorm2d(C_out),
+        ]
+        super(ShiftBlock5x5, self).__init__(*ops)
+
+    def forward(self, x):
+        y = super().forward(x)
```
fmassa: nit: `super(ShiftBlock5x5, self).forward(x)`

maskrcnn_benchmark/modeling/backbone/fbnet_builder.py
```diff
@@ -0,0 +1,822 @@
...
+
+
+class Upsample(nn.Module):
+    def __init__(self, scale_factor, mode):
+        super(Upsample, self).__init__()
+        self.scale = scale_factor
+        self.mode = mode
+
+    def forward(self, x):
+        return interpolate(x, scale_factor=self.scale, mode=self.mode)
```
fmassa: nit: I think we need to pass the `align_corners` here or else we will have a deprecation warning

configs/gn_baselines/e2e_mask_rcnn_R_50_FPN_1x_gn.yaml
```diff
@@ -1,5 +1,5 @@
 INPUT:
-  MIN_SIZE_TRAIN: 800
+  MIN_SIZE_TRAIN: (800,)
```
fmassa: Thanks for fixing those!
Quick question: the permissions of those files have been changed. Was that on purpose? Is one preferred over the other?

maskrcnn_benchmark/modeling/backbone/fbnet_builder.py
newstzpz: Updated.

maskrcnn_benchmark/modeling/backbone/fbnet_builder.py
newstzpz: Updated.

maskrcnn_benchmark/modeling/backbone/fbnet_builder.py
newstzpz: The mode we used ("nearest") requires setting align_corners=None otherwise it will throw an exception.

configs/gn_baselines/e2e_mask_rcnn_R_50_FPN_1x_gn.yaml
newstzpz: The permissions are changed incidentally. Will revert them back.

jordanxlj:
@newstzpz, can you provide fbnet pretrained model? thanks.

newstzpz:
> @newstzpz, can you provide fbnet pretrained model? thanks.

Yes, we will provide the pretrained models and benchmark numbers soon.

AnnaAraslanova:
@newstzpz Do I understand correctly, that the searched FBNet architectures (from the paper https://arxiv.org/pdf/1812.03443.pdf, page 7) was not included in maskrcnn_benchmark/modeling/backbone/fbnet_modeldef.py ?
And the file of architecture definitions (fbnet_modeldef.py) consists of examples that perform well with the detection and segmentation tasks?

PR #465, opened 2019-02-19
please set 'STRIDE_IN_1X1' to be 'False' when backbone uses GN

zimenglan-sysu-512:
to fix the performance gap compared to Detectron.


fmassa:
Can I merge this now or do you want me to wait (maybe for some results to finish computing)?

zimenglan-sysu-512:
hi @fmassa 
maybe wait for the trainings, because it will need 2 days to finish.

fmassa:
Let's do it like the following: once the trainings are over, you can send another PR updating maybe the `MODEL_ZOO` description of the GN model that you trained?

zimenglan-sysu-512:
ok.
agree with u

clw5180:
> Let's do it like the following: once the trainings are over, you can send another PR updating maybe the `MODEL_ZOO` description of the GN model that you trained?

Hi fmassa, I have a question here: What does the STRIDE_IN_1X1 mean? If I use ResNeXt101-FPN, should I set this option to True? Thanks a lot ! @fmassa

Issue #466, opened 2019-02-19
Cityscapes to COCO inefficiency


botcs:
I was looking into `tools/cityscapes/convert_cityscapes_to_coco.py` implementation, and found out that in [L167](https://github.com/facebookresearch/maskrcnn-benchmark/blob/0f9476bd2bb8cc46c5532ae5a2024a854079c5d5/tools/cityscapes/convert_cityscapes_to_coco.py#L167) it is loading a json of already computed polygons to read the metadata and find the binary mask of the corresponding entry to transfer them to polygons **AGAIN**, in [`tools/cityscapes/instances2dict_with_polygons.py`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/0f9476bd2bb8cc46c5532ae5a2024a854079c5d5/tools/cityscapes/instances2dict_with_polygons.py#L51).

Why is it necessary to recompute the polygons?

fmassa:
To be frank, I don't know.

That file was taken as is from Detectron in https://github.com/facebookresearch/maskrcnn-benchmark/pull/232 by @henrywang1 , and I didn't actually reviewed that part of the PR, given that it was taken from Detectron.

If you see optimization opportunities there, feel free to send a PR!

akshitac8:
Hello @botcs as I understood from the code it is used to read the height and width of the image. You can replace that step by cv2.imread or lycon.load to get the height and width of the desired image.
Hope this helps :slightly_smiling_face:

botcs:
I am currently refactoring the `structures/segmentation_mask.py` and once it is done, it should be a choice whether the user wants to use the binary masks or the polygons from CityScapes.

> If you see optimization opportunities there, feel free to send a PR!
Let me clarify: we download the masks and polygons both with the original dataset - and then we take the masks to generate polygons again. One possibility for this could be that this was some legacy code, maybe the polygon annotation feature was just added later. But that would contradict that the author of this piece of code already knew about the polygon annotation since it reads in the whole file just to check the width and height of the image. Thankfully this stuff goes offline but still its existence is intimidating

PR #473, opened 2019-02-21
Support Binary Mask with transparent SementationMask interface

botcs:
After a major refactor on #150 I have implemented a generic SegmentationMask with having the following in mind:

```
Segmentations come in either:
1) Binary masks
2) Polygons

Binary masks can be represented in a contiguous array
and operations can be carried out more efficiently,
therefore BinaryMaskList handles them together.

Polygons are handled separately for each instance,
by PolygonInstance and instances are handled by
PolygonList.

SegmentationMask is supposed to represent both,
therefore it wraps the functions of BinaryMaskList
and PolygonList to make it transparent.
```

## Features:
- **Error handling:** I have added multiple sanity-check at different levels of initializations, which should prevent invalid data to pass through.
- **Transparent interface - Generic backend:** `BinaryMaskList`s and `PolygonList`s are now interchangeable, and both can serve as the underlying representation of the `SegmentationMask`.
- **Unit test:** I have refactored @wangg12 's unit tests as well, and in [this notebook](https://nbviewer.jupyter.org/gist/botcs/4bd24deddc2c3c31e46601cfdd55bfe9) I have provided visual feedback on the results as well.
- **Unit train:** Quick schedule trainings were done using `configs/quick_schedules/e2e_mask_rcnn_R_50_FPN_quick.yaml` with the Binary Masks consistently outperforming (maskAP=5.8) the training using Polygon representation on `coco_minival`. Also, using the Polygon backend the similar performance (maskAP=5.6) was achieved when using the current backend in the `master` branch.

### Downsides:
I have changed the syntax to represent the structure more coherently which affects two lines in the lib: 
- [`maskrcnn_benchmark/data/datasets/coco.py`-L83]
- [`maskrcnn_benchmark/modeling/roi_heads/mask_head/loss.py`-L38]


maskrcnn_benchmark/data/datasets/coco.py
```diff
@@ -80,7 +80,7 @@ def __getitem__(self, idx):
         target.add_field("labels", classes)
 
         masks = [obj["segmentation"] for obj in anno]
-        masks = SegmentationMask(masks, img.size)
+        masks = SegmentationMask(masks, img.size, mode='poly')
```
botcs: Passing the preferred back-end is added to the argument list

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -8,68 +11,213 @@
...
+            align_corners=False
+        )[0].type_as(self.masks)
+        resized_size = width, height
+        return BinaryMaskList(resized_masks, resized_size)
+
+
+
+    def convert_to_polygon(self):
+        contours = self._findContours()
+        return PolygonList(contours, self.size)
```
wangg12: This is not generally true. Consider a  donut shape mask, the contours would fail to represent the real shape.

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -8,68 +11,213 @@
...
+                masks: Either torch.tensor of [num_instances, H, W]
+                    or list of torch.tensors of [H, W] or
+                    BinaryMaskList.
+                size: absolute image size, width first
+        '''
+
+        if isinstance(masks, torch.Tensor):
+            pass
+        elif isinstance(masks, (list, tuple)):
+            masks = torch.stack(masks, dim=2)
```
wangg12: I would like to add support for converting RLE masks to binary mask as well.

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: Please refer to [L139](https://github.com/facebookresearch/maskrcnn-benchmark/pull/473/files/d79829d0438636447e0137d9df7bcb19cf9d307d#diff-928af5178eceaef7d662fe22c85f439aR139):
```
contour, hierarchy = cv2.findContours(
  mask, cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_TC89_L1
)
```
This solves the donut problem:
- [binary mask](https://pasteboard.co/I2c1EbA2.png)
- polygon [using cv2.RETR_TREE](https://pasteboard.co/I2c2bCr.png)
- polygon [using cv2.RETR_EXTERNAL](https://pasteboard.co/I2c2Lby.png)

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: In PR #150 RLE masks are discussed, however I could not find anywhere in the code of RLE support.
In my opinion, the COCO dataset seems to go fine without RLE, and I don't see the point why would it be generally useful.

Aside from that, you can implement the `RLEList` to back up `SegmentationMask` supporting that format, without interfering with the other classes. Of course to support full compatibility, you would have to take care of the to-from conversion for both `RLEList <-> BinaryMaskList` and `RLEList <-> PolygonList`

wangg12: Again `RLE -> polygons` is not always possible considering a donut-like shape. So just convert it into binary mask in the input of binary mask class is enough.

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: Have you tested with a binary mask with multiple holes? I am not sure whether it is generalizable to all cases.

botcs: can you give a counterexample, where you cannot select the outermost contour?

wangg12: @botcs Can you convert the contours of a donut-shape object or object with holes found by cv2 back to the original binary mask via cocomask API? At least in my experience this is not trivial.

Besides, the function you provided has a minor problem with my opencv version. 
It should be `_, contour, hierarchy = cv2.findContours(
  mask, cv2.RETR_EXTERNAL, cv2.CHAIN_APPROX_TC89_L1
)`

botcs: I will check it out, could you please give me a polygon template to use?

About the opencv issue: previously I have came across an aux function that supports both versions, 
Personally, I would suggest updating it to the latest version which does not change the passed mask argument, therefore it is not returning it either since the whole project is using nightly-builds and there is little to no support on backwards compatibility.

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: Sounds reasonable, I will add it.

maskrcnn_benchmark/structures/segmentation_mask.py
wangg12: Just use the polygon testcase in my original unitest，where the second polygon is actually the hole of the object.

botcs: I removed that part from the test since it did not change anything of the binary mask, which makes me think that the conversion is performed on both polygon instances but there are no signs why would be the inner hole's mask subtracted from the outer.

One way to come around this is to convert each polygon entities of the PolygonInstance using a N-ary XOR that would translate to 1 IFF the pixel is covered by a single polygon.

What do you think?

wangg12: I think this is OK, but maybe a warning or comments should be noted to users that if their dataset contains objects with holes, this conversion is somehow problematic.

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -8,68 +11,213 @@
...
-            polygons = [torch.as_tensor(p, dtype=torch.float32) for p in polygons]
-        elif isinstance(polygons, Polygons):
-            polygons = polygons.polygons
+        else:
+            RuntimeError('Type of argument `polygons` is not allowed:%s'\
+                %(type(polygons)))
+
+        ''' This crashes the training way too many times...
+        for p in polygons:
+            assert p[::2].min() >= 0
```
fmassa: can you explain why the training crashes with those asserts? are the elements out of bounds or negatives?

maskrcnn_benchmark/modeling/roi_heads/mask_head/loss.py
```diff
@@ -27,17 +27,15 @@ def project_masks_on_boxes(segmentation_masks, proposals, discretization_size):
     assert segmentation_masks.size == proposals.size, "{}, {}".format(
         segmentation_masks, proposals
     )
-    # TODO put the proposals on the CPU, as the representation for the
```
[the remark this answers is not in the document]
fmassa: Actually I would rather have this here, but instead of a TODO as a `FIXME`, because this is currently a bottleneck in the training code

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: both cases happen in the COCO training.
Ideally, this assert should not be a problem, however it seems that when the polygons are converted to binary masks, the implementation handles them without throwing warnings.

maskrcnn_benchmark/modeling/roi_heads/mask_head/loss.py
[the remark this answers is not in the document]
botcs: Agreed.
I was thinking how could this step be speed up a bit:
```
1) Iterate
2) Crop
3) Resize
4) Stack
```
But since the `Crop` step is carried out differently for each entry, I didn't see a straightforward way to do this.
If you think that a simple Python multiprocessing thread could solve this then I would be more than happy to implement it - I am just curious if it would collide with the torch backend.

botcs:
Please also note, that now indexing in or slicing the `SegmentationMask`, `PolygonList`, `BinaryMaskList` instances returns an instance of the same class instead of the underlying representation with the corresponding size of the slice:
```
SM = SegmentationMask( ... )
type(SM[0]) == type(SM)
type(SM[1:4]) == type(SM)
```

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: I have found the implementation for RLE conversions [here](https://github.com/cocodataset/cocoapi/blob/master/PythonAPI/pycocotools/_mask.pyx), which suggested that RLEs are passed as list of dicts, each of them has at least an entry with the key `count`, so I am checking that in the initialization.

@wangg12 could you please run a test if it runs OK this way?

botcs:
I could not add a comment to @fmassa 's [comment](https://github.com/facebookresearch/maskrcnn-benchmark/pull/473#discussion_r259332526) so I will post it below:
-------------------------
>Actually I would rather have this here, but instead of a TODO as a `FIXME`, because this is currently a bottleneck in the training code

Agreed.
I was thinking how could this step be speed up a bit:

```
1) Iterate
2) Crop
3) Resize
4) Stack
```
But since the `Crop` step is carried out differently for each entry, I didn't see a straightforward way to do this.
If you think that a simple Python multiprocessing thread could solve this then I would be more than happy to implement it - I am just curious if it would collide with the torch backend.

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: I have now looked into the details of this `polygon -> mask` function carried out by COCO API, and the following happens: each polygon entity (parts of a polygon instance) is converted to RLE and the RLE is converted to mask.

This whole unnecessary overhead could be simplified by using [`cv2.fillPoly`](https://docs.opencv.org/3.0-beta/modules/imgproc/doc/drawing_functions.html?highlight=fillpoly#void%20fillPoly(Mat&%20img,%20const%20Point**%20pts,%20const%20int*%20npts,%20int%20ncontours,%20const%20Scalar&%20color,%20int%20lineType,%20int%20shift,%20Point%20offset)) which could handle the output of the `Mask->Polygon` conversion in a very neat way. This could allow a probably *lossless* `Mask<->Polygon` bi-directional conversion.

What is your opinion?

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -201,14 +440,128 @@ def __getitem__(self, item):
...
+        s += 'image_height={})'.format(self.size[1])
+        return s
+
+
+
+
+
+class SegmentationMask(object):
+
+    '''
```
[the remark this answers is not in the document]
botcs: used `black` lib to check the code

botcs:
> I have a few comments. I'd need to run a few experiments to validate that there is no change in behavior, so it might take a bit more time to merge this PR than usual.
Hi @fmassa, 
How are you going to validate exactly the consistency? I can make a notebook on comparing the old and the new implementation (both elementwise check on mask tensors and comments on syntax), if that helps.

Thank you

maskrcnn_benchmark/structures/segmentation_mask.py
fmassa: I still don't understand this correctly. This assumes that `p` is a polygon, right?

fmassa:
Hi @botcs sorry for the delay

About your first message:
This PR writes a cuda kernel that replaces the Crop / Resize / Stack https://github.com/facebookresearch/maskrcnn-benchmark/pull/379

I was maybe hoping to have a way of decomposing a bit more the structures, but this might not be possible / needed.

About validating the consistency, my thought was to run a full training and compare end performances. This would indicate if the representation that we use brings any noticeable difference in the final model result.

maskrcnn_benchmark/structures/segmentation_mask.py
fmassa: @botcs I'd prefer not to have a dependency on OpenCV in the core training loop (I've seen reports in the past that it didn't mixed well with pytorch multiprocessing, even though it might not apply here).

About lossless conversion from Mask to Polygon, I don't think this is actually possible, because there are discretization artifacts introduced when we go from a Polygon to a Mask that can't be exactly recovered. That being said, it would definitely be nice to have a way to convert back from mask to polygon, but in this case, given that polygons take less space, wouldn't it be recommended to (almost) always use polygons instead?

botcs:
What would you think about a [CityScapes](https://www.cityscapes-dataset.com/) training, where we could compare the `old-poly`, `new-poly`, `new-mask` annotation?

By `old-poly`, I mean the implementation on `master`, using coco-style training (issue filed [here](https://github.com/facebookresearch/maskrcnn-benchmark/issues/466)) with the [`configs/cityscapes/e2e_mask_rcnn_R_50_FPN_1x_cocostyle.yaml`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/configs/cityscapes/e2e_mask_rcnn_R_50_FPN_1x_cocostyle.yaml) config file

For the `new-poly` and `new-mask` I would make a `CityScapesDataset` that would take either the **original** polygons, binary masks respectively as ground-truth.

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: @fmassa I agree with the dependency issue. Should I import it only when the conversion is called and raise a warning, or remove the feature?

I agree, but apart from corner cases, in worst case the boundary pixels' coordinates would serve as the polygon representation. In reverse, if no shrinking or warping has been applied I think conversion is completely reversible.

> given that polygons take less space, wouldn't it be recommended to (almost) always use polygons instead?

in theory yes, in practice I think it depends how often does the conversion happen. In synthetic data there are ridiculously rough surfaces which would be expensive to store and as long as polygon operations are not parallelized, masks **could be** dealt with more easily. 

To sum up: In the current version polygon conversion returns the outer boundary only, no polygons with holes can be represented; on the other side we don't have parallelized binary mask transformations.

fmassa:
@botcs that sounds amazing.
Having a kind of official implementation that supports masks would be great.
But the complication is around testing code: we currently use COCO for evaluating everything, and we should stick with it or rewrite the evaluation logic (some of which is in pascal eval), but which supports masks.
But that's a lot of work.

wangg12:
@fmassa Why would evaluation be a problem since we could save the testing results in COCO style?

fmassa:
@wangg12 yes, you are right, the predictions are masks anyway and we convert them to RLE for COCO evaluation. But the COCOeval expects polygons for the ground truth, so those polygons should be there as well.

But the current way we dispatch to the evaluation code is for now hard-coded to dispatch to either the pascal evaluation or the coco evaluation, depending on the dataset class.

Those are not blockers, just problems that we need to address in order to make the code easy to use, and would definitely be a great contribution!

wangg12:
@fmassa I think if users use binary masks, they could convert their annotations losslessly to COCO-style RLE formats, so there is no need to change the evaluation code.

botcs:
I am working on a generic coco-style evaluation that is compatible (as much as possible) with segmentation datasets. However, it would definitely help if this PR would be merged, since that refactoring would be out of scope.

fmassa:
@wangg12 sounds good.

One thing I'd love to do is to have a kind of generic way of calling into the COCOeval without requiring the data to be in the COCODataset format.

wangg12:
@fmassa I agree with the need of a generic way, but converting to COCO format is currently an easy and workable solution.

botcs:
@wangg12 I would be strongly against that. No one else than me wants results out of different datasets at the moment than me, but these kind of hacks increase the technical debt and lead to monsters like #466.

As mentioned in my [previous comment](https://github.com/facebookresearch/maskrcnn-benchmark/pull/473#issuecomment-467421211), I would be happy to implement the COCOeval without hard-wiring it to COCODataset. That update however would require that first I implement a CityScapesDataset (and others as well maybe) that share a common interface - which would be built on top of the subject of the current PR, the new SegmentationMask. My concern is that if I start to merge in this PR all these patches than it would be quite untraceable what happened.

I am aware that @fmassa needs a full training to validate this PR, which could be done using the current COCO tools (this PR is compatible with the original COCOeval and COCODataset format) - that would delay the exhaustive comparison mentioned in [this comment](https://github.com/facebookresearch/maskrcnn-benchmark/pull/473#issuecomment-467170547), but could make up a good roadmap to follow during the week:

Roadmap
----
1. **[Done]** refactor `SegmentationMask`
    - **[Done]** refactor training if necessary
    - **[Done]** unit tests and *quick_schedule* training comparison
    - a single full training on COCO using a specific config file, knowing that their expected performance is already reported.
    - merge this PR
1. refactor `COCOeval` to use a generic `Dataset` class
    - refactor `COCODataset` if necessary
    - test `COCOeval` and `COCODataset` if it gives the same result for a pre-trained model as the original implementation
    - merge this PR
1. implement `CityScapesDataset`
    - test and compare full training results with using ground truth `PolygonList` and `BinaryMaskList`
    - merge this PR

maskrcnn_benchmark/structures/segmentation_mask.py
botcs: @fmassa please take a look at the `PolygonInstance.crop` method, which is a bit odd to me: instead of clamping (as noted in the in-line comment) it just subtracts the left/top boundrary of the cropping box of the polygons, and passes the reduced cropping size as an argument for the new `PolygonInstance` instance.

This was in the original implementation, and I suppose it heavily relies on the fact that `polygon->binary` conversion will handle this issue by just ignoring over- and under-flowing polygon coordinates. So this could be why it is working without, and crashing with these asserts

wangg12:
@botcs Actually I am using RLE format COCO-style annotations for the binary masks which were orinally provided by the dataset. Apart from the losslessly conversion code (which is pretty easy) and the conversion to binary masks in the data loading part, there is no need to change the other part of the codebase. 

In my opinion, in order to make the feature in this PR to work, there is no need to change the evaluation code. I'd appreciate it if a more generic  evaluation API could be supported, however that could happen in another independent PR. 😄😄😄

botcs:
@wangg12 Yeah, I agree on this, that is why point 1) has no plans on the evaluation code:
>  to make the feature in this PR to work, there is no need to change the evaluation code

Still I think it is important to know the performance differences, if there are any, caused by different representations. That is why the CityScapes would make a perfect target

fmassa:
@botcs I agree on your plan from https://github.com/facebookresearch/maskrcnn-benchmark/pull/473#issuecomment-467474884

Let me launch some trainings now with your PR to compare if there are any noticeable differences after training. I'll update you tomorrow

maskrcnn_benchmark/structures/segmentation_mask.py
fmassa: Oh I see, yes, it makes sense.

The cropping was implemented that way to mimic the implementation in Detectron for the fused `project_on_box`. Actually, to make things really correct we would need to identify that a point is negative, and add new edges to the polygon. Just clamping the coordinates of the polygon is unfortunately not enough

botcs:
@fmassa I have run two trainings using `configs/e2e_mask_rcnn_R_50_FPN_1x.yaml`, but I have **extremely strange** results. Please find the logs attached.

- One is from the master branch, without any modification, in a fresh conda env, in which the demo was working with the pretrained weights: https://gist.github.com/botcs/0ba9bab0950fdad26ff39a271fbc988f

- The other is from the PR #473, in which a batch overfitting (using only random 4 samples of the training data) works properly: https://gist.github.com/botcs/c26e92734b4e4a9da3c1eb1d51c46335

**both** of them got NaNs during training. The timeline was the following:
1. `#473` gets NaN at iteration 8200 (log file overwritten)
2. rerun `#473`, gets NaN at iteration 2600 ([log file](https://gist.github.com/botcs/c26e92734b4e4a9da3c1eb1d51c46335))
3. run `master`, for sanity check, gets NaN at iteration 14500 ([log file](https://gist.github.com/botcs/0ba9bab0950fdad26ff39a271fbc988f))

2x NVidia 2080 were used, Ubuntu 16.04, CUDA 9.
I would like to emphasize that every other hyperparameter was the very same, the modifications for #473 are the same as they are in sync with the github pr.

My biggest concern is that the training is not working on master.

fmassa:
@botcs I just run a quick training using your branch in here (without modifications) and it works fine.
I had a look at your log, and it looks like you changed the batch size but didn't update the learning rates nor the learning rate schedules?

fmassa:
More precisely, this here indicates that there is a problem with the config file https://gist.github.com/botcs/0ba9bab0950fdad26ff39a271fbc988f#file-log-master-txt-L210-L215
If you are using 2 GPUs and a smaller batch size, you should adapt the learning rate / etc following the information in https://github.com/facebookresearch/maskrcnn-benchmark#single-gpu-training

botcs:
OH sorry.. that is completely, true, I have reduced the batch size, because the 2080 has only 8GB...
Wow. my bad. extremely sorry. will update them accordingly

fmassa:
@botcs I just got a similar problem as @IssamLaradji
 in https://github.com/facebookresearch/maskrcnn-benchmark/pull/150#issuecomment-467860197 when I tried to run a quick training with binary masks:

```
    return torch._C._nn.upsample_bilinear2d(input, _output_size(2), align_corners)
RuntimeError: invalid argument 2: non-empty 4D input tensor expected but got: [1 x 0 x 640 x 480] at /opt/conda/conda-bld/pytorch-nightly_1549566624064/work/aten/src/THNN/generic/SpatialUpSamplingBilinear.c:22
```

I think this has to do with returning 0 channels in https://github.com/facebookresearch/maskrcnn-benchmark/pull/473/files#diff-928af5178eceaef7d662fe22c85f439aR403, which is not handled properly by `interpolate`.

Also, might be good using `interpolate` from https://github.com/facebookresearch/maskrcnn-benchmark/blob/75523a75312a7c8a9a007f75375f6310b8d37ad9/maskrcnn_benchmark/layers/misc.py#L75-L110 , which supports 0-size batch

botcs:
@fmassa I have patched the interpolate function, and now it used your suggested interpolate function

Should I add this case to unit test?

botcs:
Hi @fmassa,
I have finished training the `R-50-FPN-MASK` network, now with the correct solver arguments:
```
SOLVER:
  BASE_LR: 0.005
  WEIGHT_DECAY: 0.0001
  STEPS: (120000, 160000)
  MAX_ITER: 180000
  IMS_PER_BATCH: 4
```
The resulting scores are the following:
bbox mAP(.5:.95) = 0.355 (reported score is **higher** 37.8)
mask mAP(.5:.95) = 0.325 (reported score is **higher** 34.2)

Is this difference occurs because of the 2 GPU - smaller batch size - smaller learning rate or because of a significantly different behavior using the new `SegmentationMask` interface?

A detailed summary of the precision and recall:
```
Average Precision  (AP) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.355
 Average Precision  (AP) @[ IoU=0.50      | area=   all | maxDets=100 ] = 0.571
 Average Precision  (AP) @[ IoU=0.75      | area=   all | maxDets=100 ] = 0.384
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.205
 Average Precision  (AP) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.385
 Average Precision  (AP) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.464
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=  1 ] = 0.298
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets= 10 ] = 0.470
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=   all | maxDets=100 ] = 0.494
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= small | maxDets=100 ] = 0.307
 Average Recall     (AR) @[ IoU=0.50:0.95 | area=medium | maxDets=100 ] = 0.530
 Average Recall     (AR) @[ IoU=0.50:0.95 | area= large | maxDets=100 ] = 0.625
```

botcs:
[loosely related]
Hi @fmassa,

I have made good progress with the plans on a new dataset agnostic COCO-style evaluation approach, however your help would be really appreciated. Please find my results visualized at:
https://nbviewer.jupyter.org/gist/botcs/268881786683299db52d79ac1d5a9c5f

thanks,
Csabi

botcs:
[COCOEval related]
Hi @fmassa,

I have made good progress with the plans on a new dataset agnostic COCO-style evaluation approach, however your help would be really appreciated. Please find my results visualized at:
https://nbviewer.jupyter.org/gist/botcs/904d96ebb50708f7690f52c831e76018

thanks,
Csabi

botcs:
Hi @fmassa,

Are there any remaining to-dos to merge or reject this PR?

BTW, I am now doing the 3rd point in the [roadmap](https://github.com/facebookresearch/maskrcnn-benchmark/pull/473#issuecomment-467474884), and I would be happy to submit the PR from the 2nd point.

fmassa:
Hey @botcs , sorry for the delay in replying.

> Is this difference occurs because of the 2 GPU - smaller batch size - smaller learning rate or because of a significantly different behavior using the new SegmentationMask interface?

This seems to indicate that the new `SegmentationMask` might give lower results than the previous one using polygons. I'll try running the training on 8 GPUs again to verify, and I'll let you know tomorrow what results I got.

> I have made good progress with the plans on a new dataset agnostic COCO-style evaluation approach, however your help would be really appreciated. Please find my results visualized at:
> BTW, I am now doing the 3rd point in the roadmap, and I would be happy to submit the PR from the 2nd point.

I had a quick look and it looked very interesting and promising. I'll have a closer look tomorrow make some more comments (I need to run now).

fmassa:
@botcs I've launched two trainings yesterday with 8GPUs each for Mask R-CNN R-50 FPN. They are still training and the time per iteration is at about 2.2915 s, which is 5x slower than the equivalent implementation using Polygons.

Does that match your expectations?

botcs:
@fmassa I have experienced a 2x slower computation, but this is because of the CPU intensive operations, probably the config is different.

If you use the polygon representation the speed is expected to be the same, also my current training on the CityScapes polygon annotations can confirm this: time: 0.4781 (0.4750)  data: 0.0091 (0.0093) for the original image size (1024x2048).

The discussion in #527 was mainly motivated by the increased process time... at the moment no tricks and heuristics are applied, just the original `mask - poly` representation refactored for the `SegmentationMask`.

wangg12:
@fmassa I've experienced a training speed with about 1s/iter by using [mmdetection](https://github.com/open-mmlab/mmdetection) and feeding the annotations with RLE based segmentation which are then converted into binary masks for training. So I am expecting that the speed should be faster with this repo.

botcs:
@wangg12 I agree.

We can speed up here literally every for loop by simply just multithreading them. Which could save a lot of time. Also the overall spectacular training speed of `mmdetection` using binary masks can depend on multiple factors as well (not even talking about the image size), as I mentioned in the previous comment, the `mask_head` does not utilize any advantages of having the segmentations in a single torch tensor.

But to be fair.. the original implementation of the `SegmentationMask` should perform similarly in time complexity to this PR's, but my point is that the primary goal of this PR was not optimizing the binary mask processing, but unifying the interface for polygons and binary masks.

fmassa:
@botcs I've finished training with masks, but evaluation failed with the same type of error as before (the 0-d tensor)
```
RuntimeError: invalid argument 2: non-empty 4D input tensor expected but got: [1 x 0 x 480 x 640] at /opt/conda/conda-bld/pytorch-nightly_1551330564688/work/aten/src/THNN/generic/SpatialUpSamplingBilinear.c:22
```
so the fixes are not yet enough for this to be merged. :-/

botcs:
@fmassa I am sad to hear that, I thought that the [interpolation commit](https://github.com/facebookresearch/maskrcnn-benchmark/pull/473/commits/493e3f6a1b660c212933e24cdb8a7c217c7dc45b) has fixed this issue. Can you please send me the full error-stack? Is it possible that a slicing operation is called after extracting the internal data representation by `.get_mask_tensor()`?

botcs:
@fmassa in the COCO evaluation script found at `maskrcnn_benchmark/data/datasets/evaluation/coco/coco_eval.py`, there is no occurrence of the `SegmentationMask` class. The process involves conversion **to** RLE **from** the `prediction.get_field("mask")`, which is the torch tensor itself (not an instance of the `SegmentationMask`, on this issue see #527 for full details). The error reported may be related to the original `Masker` implementation found at `maskrcnn_benchmark/modeling/roi_heads/mask_head/inference.py`, where the `interpolation` function is imported from `torch.nn.functional` instead of `maskrcnn_benchmark.layers.misc` - the same problem that caused error in this implementation.

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -8,63 +11,207 @@
...
+        if isinstance(masks, torch.Tensor):
+            # The raw data representation is passed as argument
+            masks = masks.clone()
+        elif isinstance(masks, (list, tuple)):
+            if isinstance(masks[0], torch.Tensor):
+                masks = torch.stack(masks, dim=2).clone()
+            elif isinstance(masks[0], dict) and "count" in masks[0]:
+                # RLE interpretation
+
+                masks = mask_utils
```
IssamLaradji: Shouldn't this be different for the case of a list of RLEs?

fmassa: cc @botcs

botcs: @IssamLaradji Nice catch!

Here is a quick guide for usage (and for sanity check on update):
https://gist.github.com/botcs/a59f3f59e22e5df93e3e5e4f86718af3

botcs: Fixed in PR #657

IssamLaradji: Perfect! the updated version looks good!

IssamLaradji: Quick question, will this get merged into master :P?

botcs: :P cc @fmassa can you merge #657?

Thanks

PR #475, opened 2019-02-21
let the saved predictions.pth make more sense

CoinCheung:
Maybe this will be helpful if we want to match the predictions with the associated image_ids.

### DIFF
--- a/maskrcnn_benchmark/data/datasets/coco.py
+++ b/maskrcnn_benchmark/data/datasets/coco.py
@@ -93,7 +93,7 @@ class COCODataset(torchvision.datasets.coco.CocoDetection):
         if self.transforms is not None:
             img, target = self.transforms(img, target)
 
-        return img, target, idx
+        return img, target, self.ids[idx]
 
     def get_img_info(self, index):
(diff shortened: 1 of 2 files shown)


fmassa:
Hi,

While this makes sense, it makes it kind of incompatible with the Pascal dataset.

Also, the meaning of `predictions.pth` is now the same for both datasets: the index is wrt to index of the dataset, and this PR would change it to use the numeric indices from coco itself.

I think that documenting this behavior would be maybe better, and would avoid backward-incompatible files with the codebase?

CoinCheung:
Thanks a lot for explaining. So after I have run my test script, how could I use the results to draw the predictions to my test set images?

CoinCheung:
I am closing this since this is not the expected way to modify the code :)

fmassa:
Hi,

Sorry for the delay in replying.

> how could I use the results to draw the predictions to my test set images?

So, I would load the `predictions.pth` file and the `COCODataset` corresponding to it, and then we are able to get the mapping from indices to image ids

CoinCheung:
Got it, I have try it. Thanks!!

PR #510, opened 2019-02-28
fix INSTALL.md

bmetge:
-> don't need the pytorch nightly
-> not enforcing cuda 9.0
-> using torchvision
-> reorder so the pip install -r requirement.txt comes after clone and not before

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -23,8 +23,6 @@ conda activate maskrcnn_benchmark
 # this installs the right pip and dependencies for the fresh python
 conda install ipython
 
-conda install -c pytorch pytorch torchvision cudatoolkit=9.0
-
 export INSTALL_DIR=$PWD
 
 
@@ -41,6 +39,10 @@ cd maskrcnn-benchmark
 
 # maskrcnn_benchmark and coco api dependencies
 pip install -r requirements.txt
+
+# follow PyTorch installation in https://pytorch.org/get-started/locally/
+# we give the instructions for CUDA 9.0
+conda install -c pytorch pytorch torchvision cudatoolkit=9.0
 
 # the following will install the lib with
 # symbolic links, so that you can modify
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -20,22 +20,10 @@
 conda create --name maskrcnn_benchmark
 conda activate maskrcnn_benchmark
 
-# this installs the right pip and dependencies for the fresh python
-conda install ipython
-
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -20,22 +20,10 @@
...
-# this installs the right pip and dependencies for the fresh python
-conda install ipython
-
-# maskrcnn_benchmark and coco api dependencies
-pip install -r requirements.txt
-
-# follow PyTorch installation in https://pytorch.org/get-started/locally/
-# we give the instructions for CUDA 9.0
-conda install pytorch-nightly cudatoolkit=9.0 -c pytorch
+conda install -c pytorch pytorch torchvision opencv
```
fmassa: Don't we need the instructions for cuda now that we have changed the way it is setup?
Also, is opencv a hard-requirement? I don't think so, except for the demos, right?

INSTALL.md
```diff
@@ -20,22 +20,10 @@
 conda create --name maskrcnn_benchmark
 conda activate maskrcnn_benchmark
 
-# this installs the right pip and dependencies for the fresh python
-conda install ipython
```
fmassa: Please let's keep this `ipython`, it's always better to have it or else `pip` might point to the wrong `pip`

INSTALL.md
bmetge: yes indeed

INSTALL.md
```diff
@@ -23,19 +23,10 @@ conda activate maskrcnn_benchmark
 # this installs the right pip and dependencies for the fresh python
 conda install ipython
 
-# maskrcnn_benchmark and coco api dependencies
-pip install -r requirements.txt
-
-# follow PyTorch installation in https://pytorch.org/get-started/locally/
```
fmassa: Can you keep this comment? This is very helpful for beginners, specially if they want to change the CUDA version

chengyangfu: I also vote for keeping this part.  Personally, I like to compile PyTorch by myself instead of installing from conda/pip.

bmetge: Makes sense indeed. Just put it back.

estimoelmo:
had a problem about the pycoco stuff and need to install `pip3 install Cython` first.

fmassa:
@estimoelmo yes, you are right, this patch here is broken because `pycocotools` require cython.

I'll revert those changes, and update the readme to point to pytorch nightly

PR #537, opened 2019-03-06
Support for running on arbitrary CUDA device.

atranitell:
The current version could not normally run someone CUDA device (e.g. **cuda:1**).
The program will show such errors if we only add a extra flag `MODEL.DEVICE cuda:1`
`
2019-03-06 19:41:24,027 maskrcnn_benchmark.trainer INFO: Start training
THCudaCheck FAIL file=/home/dl/KJ/dl/core/maskrcnn-benchmark/maskrcnn_benchmark/csrc/cuda/nms.cu line=103 error=77 : an illegal memory access was encountered
`
The issue is that the current executable device (on CUDA) is device:0, while the memory data of tensors (boxes, etc.) is allocated on device:1. Therefore, we need temporary switch device 0 to 1.

After fixed the issue:
`2019-03-06 19:51:47,698 maskrcnn_benchmark.trainer INFO: Start training
2019-03-06 19:51:55,449 maskrcnn_benchmark.trainer INFO: eta: 3 days, 5:29:26  iter: 20  loss: 1.8457 (2.2689)  loss_classifier: 0.4143 (0.8737)  loss_box_reg: 0.0372 (0.0519)  loss_mask: 0.7941 (0.8378)  loss_objectness: 0.3711 (0.4117)  loss_rpn_box_reg: 0.0559 (0.0938)  time: 0.3500 (0.3875)  data: 0.0087 (0.0445)  lr: 0.001793  max mem: 3362
2019-03-06 19:52:02,983 maskrcnn_benchmark.trainer INFO: eta: 3 days, 4:24:40  iter: 40  loss: 1.3515 (1.9323)  loss_classifier: 0.3346 (0.6586)  loss_box_reg: 0.0855 (0.0689)  loss_mask: 0.7014 (0.7690)  loss_objectness: 0.2152 (0.3520)  loss_rpn_box_reg: 0.0460 (0.0838)  time: 0.3729 (0.3821)  data: 0.0100 (0.0277)  lr: 0.001927  max mem: 3466
`
The support could help some developer run the program in some limited environments.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,16 +70,24 @@ predictions = coco_demo.run_on_opencv_image(image)
 
 ### Use it on an arbitrary GPU device
 For some cases, while multi-GPU devices are installed in a machine, a possible situation is that 
-we only have accesse to a specified GPU device (e.g. CUDA:1 or CUDA:2) for inference, testing or training.
+we only have accesse to a specified GPU device (e.g. CUDA:1 or CUDA:2) for inference, testing or training. 
+Here, the repository currently supports two methods to control devices.
 
+#### 1. using CUDA_VISIBLE_DEVICES environment variable (Recommend)
 Here is an example for Mask R-CNN R-50 FPN quick on the second device (CUDA:1):
 ```bash
-# for training
-python tools/train_net.py --config-file=configs/quick_schedules/e2e_mask_rcnn_R_50_FPN_quick.yaml MODEL.DEVICE cuda:1
-# for testing
-python tools/test_net.py --config-file=configs/quick_schedules/e2e_mask_rcnn_R_50_FPN_quick.yaml MODEL.DEVICE cuda:1
+export CUDA_VISIBLE_DEVICES=1
+python tools/train_net.py --config-file=configs/quick_schedules/e2e_mask_rcnn_R_50_FPN_quick.yaml
 ```
-Where, we add a `MODEL.DEVICE cuda:1` flag to specify the target device running the program.
+Now, the session will be totally loaded on the second GPU device (CUDA:1).
+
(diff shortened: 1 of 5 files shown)


atranitell:
Hi Massa,
The CUDAGuard is indeed an elegant method to maintain the executive devices.
I have added it to the commit.
Best,
Kai J

jario-jin:
It is better to add a description in README.md, I also want to run on arbitrary CUDA device, but still don't know how to use. Thanks

atranitell:
> It is better to add a description in README.md, I also want to run on arbitrary CUDA device, but still don't know how to use. Thanks

Yeah, it's necessary to tell the usage for users.

README.md
```diff
@@ -68,6 +68,19 @@ image = ...
...
 ```
 
+### Use it on an arbitrary GPU device
+For some cases, while multi-GPU devices are installed in a machine, a possible situation is that 
+we only have accesse to a specified GPU device (e.g. CUDA:1 or CUDA:2) for inference, testing or training.
+
+Here is an example for Mask R-CNN R-50 FPN quick on the second device (CUDA:1):
+```bash
+# for training
+python tools/train_net.py --config-file=configs/quick_schedules/e2e_mask_rcnn_R_50_FPN_quick.yaml MODEL.DEVICE cuda:1
```
fmassa: I would not recommend this as the first method for running on the a different GPU.

Indeed, this used to still allocate some memory on the first GPU. Nowadays this is still the case but much less (~10MB from what I tried), while using `CUDA_VISIBLE_DEVICES` really doesn't use anything.

After you make this change and add first a note on this env var then I think this PR is good to go

PR #543, opened 2019-03-07
Update ImageList to work with 3d tensors

fmassa:
Useful for some examples with the webcam


dannyhung1128:
Hi, do we need custom 3D nms / roi pooling / roi align when our input data is 3 dimensional? Thanks

fmassa:
probably

PR #555, opened 2019-03-11
Fix dispatch breakage

vishwakftw:
@fmassa This fixes the dispatch breakage introduced by https://github.com/pytorch/pytorch/commit/3aeb78079bcd68282fe9117088e138b77318e288.

Closes https://github.com/pytorch/pytorch/issues/17861

### DIFF
--- a/maskrcnn_benchmark/csrc/cpu/ROIAlign_cpu.cpp
+++ b/maskrcnn_benchmark/csrc/cpu/ROIAlign_cpu.cpp
@@ -239,7 +239,7 @@ at::Tensor ROIAlign_forward_cpu(const at::Tensor& input,
     return output;
   }
 
-  AT_DISPATCH_FLOATING_TYPES(input.type(), "ROIAlign_forward", [&] {
+  AT_DISPATCH_FLOATING_TYPES(input.scalar_type(), "ROIAlign_forward", [&] {
     ROIAlignForward_cpu_kernel<scalar_t>(
          output_size,
          input.data<scalar_t>(),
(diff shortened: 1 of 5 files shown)


fmassa:
Thanks!

But I suppose that we need now to ask users to use PyTorch from source, or else it will break for those who are using PyTorch 1.0, or 1.0.1, right?

vishwakftw:
The installation instructions suggested to use the nightly build, which is why I sent in a PR.

fmassa:
That used to be the case, but we have since https://github.com/facebookresearch/maskrcnn-benchmark/pull/510 changed it to install from a pinned version.

Something should be changed, here are a few possibilities:
- change the install instructions to install from a nightly, and add a big warning there saying that this repo doesn't work with 1.0 or 1.0.1
- revert this patch and keep this repo only working on a pinned version of PyTorch
- find a better solution

fmassa:
As it turns out, #510 had a bugged install anyway, so I'll be reverting parts of it, and in the meantime I'll also make it work only on a nightly version

vishwakftw:
Oh I'm sorry I didn't see the installation snippet. The first line in [INSTALL.md](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/INSTALL.md) seems to be misleading.

fmassa:
@vishwakftw I'll fix the installation instructions, don't worry, they are bugged now anyway :-D

vishwakftw:
Thank you, and sorry about the confusion.

fmassa:
No worries, thanks for fixing this issue so quickly!

drcege:
> @vishwakftw I'll fix the installation instructions, don't worry, they are bugged now anyway :-D
 
The first line of Installation/Requirements (`PyTorch 1.0 from a nightly release`) is still misleading.
The nightly release is now `PyTorch 1.1` (am I right?).  And README also said the project is using PyTorch 1.0.

By the way, are Dockerfiles up to date?

Issue #521, opened 2019-03-01
Step-by-step tutorial - How to train your own dataset


AdanMora:
## 🚀 Feature: Training a custom dataset

In order to complete and unify all the issues about how to train a model using a custom dataset, you will find where the minimal steps to do it. Taking into account that this implementation it's a recent release, this steps can change in the future. I'm requesting feedback, this was the steps that I followed for training models and it works perfectly.

In my case, the context of my problem is completely different to the COCO dataset, with only 4 classes.

Related issues -> #15, #297, etc.

### 1) COCO format

The easier way to to use this model is labelling your dataset with the [official coco format ](http://cocodataset.org/#format-data) according with the problem you have. In my case, It's for instance segmentation, so I used the detection format.

### 2) Creating a `Dataset` class for your data

Following the example [`coco.py`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/data/datasets/coco.py). Create a new class extending from [`torchvision.datasets.coco.CocoDetection`](https://pytorch.org/docs/0.4.0/_modules/torchvision/datasets/coco.html) (you can find another classes in the official docs), this class encapsulates the pycocoapi methods to manage your coco dataset.

This class has to be created in `maskrcnn-benchmark/maskrcnn_benchmark/data/datasets` folder, same as `coco.py`, and included in the `__init__.py`.

### 3) Adding dataset paths

This class will needs as parameters the paths for the `JSON` file, it contains the metadata of your dataset in coco format, and for the images, the folder where they are. The engine automatically searches in [`paths_catalog.py`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/config/paths_catalog.py) for these parameters, the easier way is including your paths to the dict `DATASETS` following the format of course, then include an `elif` statement in the `get` method.

### 4) Evaluation file

Here is the importance of use the coco format, if your dataset have the same structure then you can use the same evaluation file used for the class `COCODataset`, in the [`__init_.py`](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/data/datasets/evaluation/__init__.py) file just add an if statement like the `COCODataset` statement.

This evaluation file follows the coco evaluation standard with the pycocoapi evaluation methods. You can create your own evaluation file, and you have to do it if your dataset have another structure.

### 5) Training script

[...]

PR #569, opened 2019-03-14
README - add your own dataset

bernhardschaefer:
Documents the places where the python code needs to be changed in order to add a custom dataset.

Proposing this to reduce the volume of issues until there is a more extensive tutorial as started in https://github.com/facebookresearch/maskrcnn-benchmark/issues/521.


bernhardschaefer:
(Reopen after accidental close)

Issue #167, opened 2018-11-17
Docker Runtime Error: Not Compiled with GPU support


archdyn:
## ❓ Questions and Help
Hello,

i have a strange Problem with the Docker Image. When I build the Docker Image given the instructions in INSTALL.md and if I then try training on the coco2014 dataset with the command below I get RuntimeError: Not compiled with GPU support(nms at ./maskrcnn_benchmakr/csrc/nms.h:22)

```
nvidia-docker run --shm-size=8gb -v /home/archdyn/Datasets/coco:/maskrcnn-benchmark/datasets/coco maskrcnn-benchmark python /maskrcnn-benchmark/tools/train_net.py --config-file "/maskrcnn-benchmark/configs/e2e_mask_rcnn_R_50_FPN_1x.yaml" SOLVER.IMS_PER_BATCH 1 SOLVER.BASE_LR 0.0025 SOLVER.MAX_ITER 720000 SOLVER.STEPS "(480000, 640000)" TEST.IMS_PER_BATCH 1
```

But whhen I change the Dockerfile and comment the line `python setup.py build develop` before `WORKDIR /maskrcnn-benchmark` out and then execute the line `python setup.py build develop` inside my built docker container i can train without problems.

My Environment when running the Docker Container:
```
2018-11-17 20:03:13,889 maskrcnn_benchmark INFO: Collecting env info (might take some time)
2018-11-17 20:03:15,634 maskrcnn_benchmark INFO: 
PyTorch version: 1.0.0.dev20181116
Is debug build: No
CUDA used to build PyTorch: 9.0.176

OS: Ubuntu 16.04.5 LTS
GCC version: (Ubuntu 5.4.0-6ubuntu1~16.04.10) 5.4.0 20160609
CMake version: version 3.5.1

Python version: 3.6
Is CUDA available: Yes
CUDA runtime version: 9.0.176
GPU models and configuration: GPU 0: GeForce GTX 850M
Nvidia driver version: 410.73
cuDNN version: Probably one of the following:
/usr/lib/x86_64-linux-gnu/libcudnn.so.7.4.1
/usr/lib/x86_64-linux-gnu/libcudnn_static_v7.a

Versions of relevant libraries:
[pip] Could not collect
[conda] pytorch-nightly           1.0.0.dev20181116 py3.6_cuda9.0.176_cudnn7.1.2_0    pytorch
        Pillow (5.3.0)
```

Does somebody know why this problem happens?

[...]

PR #572, opened 2019-03-15
Remove torch.cuda check in setup.py

eskjorg:
Let's say you want to run `docker build` on a machine without a GPU (for later use on a machine with GPU). In that case, only the `CUDA_HOME is not None` should be enough if CUDA is installed in the base image.

`torch.cuda.is_available()` on the other hand would evaluate to `false`. Or is there any important reason keep that check?

See hack at:
https://github.com/facebookresearch/maskrcnn-benchmark/issues/167#issuecomment-452796140

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -28,7 +28,7 @@ def get_extensions():
     extra_compile_args = {"cxx": []}
     define_macros = []
 
-    if torch.cuda.is_available() and CUDA_HOME is not None:
+    if CUDA_HOME is not None:
         extension = CUDAExtension
         sources += source_cuda
         define_macros += [("WITH_CUDA", None)]


fmassa:
Thanks for the PR, but I'm afraid that this will cause problems for a number of other users, so I'd rather not bias towards this use-case I think.

eskjorg:
> Thanks for the PR, but I'm afraid that this will cause problems for a number of other users, so I'd rather not bias towards this use-case I think.

Just out of curiosity, do you know any examples in which problems could arise?

fmassa:
Say you have some bits of the CUDA installed in your machine (even if you don't have GPUs). That was a requirement for Caffe to install if I remember properly.
Now you'd try to compile the extensions and I'm really not sure this would work out fine or give weird problems to CPU-only users.

The docker setup is an optional setup in maskrcnn-benchmark. The recommended installation instructions are by using conda, and this has worked out well so far, so I'd rather not change it.

miguelvr:
The PR in #612 should fix the issue.

eskjorg:
Great! Closing this.

PR #612, opened 2019-03-27
add the option to use a `FORCE_CUDA` to force cuda installation on docker

bendidi:
cf [#167](https://github.com/facebookresearch/maskrcnn-benchmark/issues/167) 

this seems to be the best workaround

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -2,10 +2,10 @@
 
 This folder contains a simple webcam demo that illustrates how you can use `maskrcnn_benchmark` for inference.
 
-
 ### With your preferred environment
 
 You can start it by running it from this folder, using one of the following commands:
+
 ```bash
 # by default, it runs on the GPU
 # for best results, use min-image-size 800
@@ -22,15 +22,15 @@ python webcam.py --min-image-size 300 --show-mask-heatmaps MODEL.DEVICE cpu
 
 Build the image with the tag `maskrcnn-benchmark` (check [INSTALL.md](../INSTALL.md) for instructions)
 
-Adjust permissions of the X server host (be careful with this step, refer to 
+Adjust permissions of the X server host (be careful with this step, refer to
 [here](http://wiki.ros.org/docker/Tutorials/GUI) for alternatives)
 
 ```bash
 xhost +
-``` 
+```
 
 Then run a container with the demo:
- 
+
 ```
 docker run --rm -it \
     -e DISPLAY=${DISPLAY} \
@@ -38,8 +38,9 @@ docker run --rm -it \
     -v /tmp/.X11-unix:/tmp/.X11-unix \
     --device=/dev/video0:/dev/video0 \
     --ipc=host maskrcnn-benchmark \
-    python demo/webcam.py --min-image-size 300
+    python demo/webcam.py --min-image-size 300 \
+    --config-file configs/caffe2/e2e_mask_rcnn_R_50_FPN_1x_caffe2.yaml
 ```
 
-**DISCLAIMER:** *This was tested for an Ubuntu 16.04 machine, 
-the volume mapping may vary depending on your platform*
+**DISCLAIMER:** _This was tested for an Ubuntu 16.04 machine,
+the volume mapping may vary depending on your platform_
--- a/demo/README.md
+++ b/demo/README.md
@@ -2,10 +2,10 @@
 
 This folder contains a simple webcam demo that illustrates how you can use `maskrcnn_benchmark` for inference.
 
+
 ### With your preferred environment
 
 You can start it by running it from this folder, using one of the following commands:
-
 ```bash
 # by default, it runs on the GPU
 # for best results, use min-image-size 800
@@ -22,15 +22,15 @@ python webcam.py --min-image-size 300 --show-mask-heatmaps MODEL.DEVICE cpu
 
 Build the image with the tag `maskrcnn-benchmark` (check [INSTALL.md](../INSTALL.md) for instructions)
 
-Adjust permissions of the X server host (be careful with this step, refer to
+Adjust permissions of the X server host (be careful with this step, refer to 
 [here](http://wiki.ros.org/docker/Tutorials/GUI) for alternatives)
 
 ```bash
 xhost +
-```
+``` 
 
 Then run a container with the demo:
-
+ 
 ```
 docker run --rm -it \
     -e DISPLAY=${DISPLAY} \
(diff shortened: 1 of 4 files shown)


miguelvr:
it is better to add `FORCE_CUDA` as a docker build arg so that it is still possible to build with CPU support only

miguelvr:
You can keep it as a default set to true. 

Plus, the documentation in `INSTALL.md` needs to be updated accordingly

docker/Dockerfile
```diff
@@ -32,7 +33,6 @@ RUN conda install -y ipython
 RUN pip install ninja yacs cython matplotlib opencv-python
 
 # Install PyTorch 1.0 Nightly
-ARG CUDA
```
miguelvr: This needs to be here

docker/Dockerfile
```diff
@@ -47,8 +47,8 @@ RUN git clone https://github.com/cocodataset/cocoapi.git \
  && python setup.py build_ext install
 
 # install PyTorch Detection
-RUN git clone https://github.com/facebookresearch/maskrcnn-benchmark.git \
+RUN git clone https://github.com/bendidi/maskrcnn-benchmark.git \
```
miguelvr: ???

bendidi: just for testing, on my remote gpu

docker/Dockerfile
bendidi: fixed

docker/Dockerfile
bendidi: fixed

docker/Dockerfile
```diff
@@ -1,5 +1,6 @@
 ARG CUDA="9.0"
 ARG CUDNN="7"
+ARG FORCE_CUDA="1"
```
miguelvr: this will go out of context as soon as the `FROM` is called, so it needs to be further down the line.

docker/Dockerfile
```diff
@@ -48,8 +48,8 @@ RUN git clone https://github.com/cocodataset/cocoapi.git \
  && python setup.py build_ext install
 
 # install PyTorch Detection
-RUN git clone https://github.com/bendidi/maskrcnn-benchmark.git \
+RUN git clone https://github.com/facebookresearch/maskrcnn-benchmark.git \
  && cd maskrcnn-benchmark \
- && FORCE_CUDA=1 python setup.py build develop
```
miguelvr: please do it like this:

```
# install PyTorch Detection
ARG FORCE_CUDA="1"
ENV FORCE_CUDA=${FORCE_CUDA}
RUN git clone https://github.com/facebookresearch/maskrcnn-benchmark.git \
 && cd maskrcnn-benchmark \
 && python setup.py build develop
```

build args are limited to the scope of the build and are not environment variables, they are intended to change the behavior of the build.

miguelvr:
@bendidi can you also modify the instructions in `INSTALL.md`? Thanks!

demo/README.md
```diff
@@ -38,7 +38,8 @@ docker run --rm -it \
     -v /tmp/.X11-unix:/tmp/.X11-unix \
     --device=/dev/video0:/dev/video0 \
     --ipc=host maskrcnn-benchmark \
-    python demo/webcam.py --min-image-size 300
+    python demo/webcam.py --min-image-size 300 \
```
miguelvr: is this change necessary?

bendidi: yes the default is `../configs/caffe2/e2e_mask_rcnn_R_50_FPN_1x_caffe2.yaml` , while when running from docker and from `maskrcnn-benchmark` folder  it should be `configs/caffe2/e2e_mask_rcnn_R_50_FPN_1x_caffe2.yaml`

bendidi:
> @bendidi can you also modify the instructions in `INSTALL.md`? Thanks!

There are no modifications to be added to `install.md`, the installation instructions stays the same. this is just a fix for the installation on docker problem and doesn't impact the process

demo/README.md
miguelvr: alright, thanks!

miguelvr:
> > @bendidi can you also modify the instructions in `INSTALL.md`? Thanks!
> 
> There are no modifications to be added to `install.md`, the installation instructions stays the same. this is just a fix for the installation on docker problem and doesn't impact the process

there are changes because there is a new build arg in the docker

demo/README.md
fmassa: Note that the demo won't work if you run it from `maskrcnn-benchmark` folder due to the local includes, so this breaks in that case

fmassa: for example, trying to `import predictor` won't work if done from `maskrcnn-benchmark`, and it is explicitly mentioned in the readme that you should cd to `demo` first, so I'd vote for removing this change

bendidi: running the current command: 
```
docker run --rm -it \
    -e DISPLAY=${DISPLAY} \
    --privileged \
    -v /tmp/.X11-unix:/tmp/.X11-unix \
    --device=/dev/video0:/dev/video0 \
    --ipc=host maskrcnn-benchmark \
    python demo/webcam.py --min-image-size 300
```
throws the error `FileNotFoundError: [Errno 2] No such file or directory: '../configs/caffe2/e2e_mask_rcnn_R_50_FPN_1x_caffe2.yaml'`

this change is just to specify path to config file in this particular case: 
```
docker run --rm -it \
    -e DISPLAY=${DISPLAY} \
    --privileged \
    -v /tmp/.X11-unix:/tmp/.X11-unix \
    --device=/dev/video0:/dev/video0 \
    --ipc=host maskrcnn-benchmark \
    python demo/webcam.py --min-image-size 300 \
    --config-file configs/caffe2/e2e_mask_rcnn_R_50_FPN_1x_caffe2.yaml
```

it should not break any imports in the demo folder

bendidi:
@miguelvr @fmassa good to merge ?

miguelvr:
Okay by me, but @fmassa has to merge

demo/README.md
fmassa: I just noticed that this is for the `docker` example.

I have never tried using it with docker, but I know for sure that without docker running the webcam demo code from `maskrcnn-benchmark` folder will not work, you need to be in the `demo` folder.

But maybe some paths are setup differently in the dockerfile.

bendidi: wierd, works fine for me outside of `demo` folder (not in docker) using python 3.6 :

```
python demo/webcam.py --min-image-size 300 --config-file configs/caffe2/e2e_mask_rcnn_R_50_FPN_1x_caffe2.yaml MODEL.DEVICE cpu
```

PR #627, opened 2019-04-02
Update grouped_batch_sampler.py

botcs:
If the dataset is empty, then the group_ids will be an empty tensor, but with type of float32 (default tensor type), which will yield an odd error message.

### DIFF
--- a/maskrcnn_benchmark/data/samplers/grouped_batch_sampler.py
+++ b/maskrcnn_benchmark/data/samplers/grouped_batch_sampler.py
@@ -40,7 +40,7 @@ class GroupedBatchSampler(BatchSampler):
     def _prepare_batches(self):
         dataset_size = len(self.group_ids)
         # get the sampled indices from the sampler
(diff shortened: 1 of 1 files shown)


maskrcnn_benchmark/data/samplers/grouped_batch_sampler.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, sampler, group_ids, batch_size, drop_uneven=False):
     def _prepare_batches(self):
         dataset_size = len(self.group_ids)
         # get the sampled indices from the sampler
-        sampled_ids = torch.as_tensor(list(self.sampler))
+        sampled_ids = torch.as_tensor(list(self.sampler)).long()
```
fmassa: Could you maybe use the new API for that?
```python
torch.as_tensor(..., dtype=torch.int64)
```

botcs:
done! :)

PR #631, opened 2019-04-03
minor fix: incorrect assert string message formatting

bernhardschaefer:
The assert message is split up in 2 strings:
```python
 "SOLVER.IMS_PER_BATCH ({}) must be divisible by the number "
        "of GPUs ({}) used.".format(images_per_batch, num_gpus)
```

str.format() is only applied to the second string, which produces an incorrect message, where the first placeholder is not substituted and the second is incorrectly substituted.

Example: `SOLVER.IMS_PER_BATCH ({}) must be divisible by the number of GPUs (32) used.`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,9 @@ dist/
 # Pycharm editor settings
 .idea
 
+# MacOS
+.DS_Store
+
 # project dirs
 /datasets
 /models
--- a/maskrcnn_benchmark/data/build.py
+++ b/maskrcnn_benchmark/data/build.py
@@ -109,18 +109,18 @@ def make_data_loader(cfg, is_train=True, is_distributed=False, start_iter=0):
     if is_train:
         images_per_batch = cfg.SOLVER.IMS_PER_BATCH
         assert (
-            images_per_batch % num_gpus == 0
-        ), "SOLVER.IMS_PER_BATCH ({}) must be divisible by the number "
-        "of GPUs ({}) used.".format(images_per_batch, num_gpus)
+                images_per_batch % num_gpus == 0
+        ), "SOLVER.IMS_PER_BATCH ({}) must be divisible by the number of GPUs ({}) used.".format(
+            images_per_batch, num_gpus)
         images_per_gpu = images_per_batch // num_gpus
         shuffle = True
         num_iters = cfg.SOLVER.MAX_ITER
     else:
         images_per_batch = cfg.TEST.IMS_PER_BATCH
         assert (
-            images_per_batch % num_gpus == 0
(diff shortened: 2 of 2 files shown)


fmassa:
Are you sure that this is necessary? I tried a small example in Python 3 and it worked ok
```python
In [16]: ("a {}"
    ...:  " bb {}".format(0, 10))
    ...:
Out[16]: 'a 0 bb 10'
```

The reason why we had the string split in two was because of the linter. While I would be ok breaking the linter here, I'd like to understand why my example case works.

bernhardschaefer:
I'm not sure if this is related to how assert is being evaluated.
Here is a small example: 
```python
Python 3.7.2 (default, Dec 29 2018, 00:00:04) 
[Clang 4.0.1 (tags/RELEASE_401/final)] on darwin
In[2]: images_per_batch = 11
In[3]: num_gpus = 3
In[4]: assert (
  ...:         images_per_batch % num_gpus == 0
  ...: ), "TEST.IMS_PER_BATCH ({}) must be divisible by the number "
  ...: "of GPUs ({}) used.".format(images_per_batch, num_gpus)
  ...: 
Traceback (most recent call last):
  File "/usr/local/anaconda3/envs/sketch2bot/lib/python3.7/site-packages/IPython/core/interactiveshell.py", line 3291, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-4-38b2e4ea65b1>", line 3, in <module>
    ), "TEST.IMS_PER_BATCH ({}) must be divisible by the number "
AssertionError: TEST.IMS_PER_BATCH ({}) must be divisible by the number 
```

To show that this is not related to IPython, here is the error I had when running the train script:
```python
AssertionError: SOLVER.IMS_PER_BATCH ({}) must be divisible by the number
Traceback (most recent call last):
  File "tools/train_net.py", line 220, in <module>
    main()
  File "tools/train_net.py", line 213, in main
    model = train(cfg, args.local_rank, args.distributed)
  File "tools/train_net.py", line 62, in train
    start_iter=arguments["iteration"],
  File "/home/ubuntu/maskrcnn-benchmark/maskrcnn_benchmark/data/build.py", line 115, in make_data_loader
    ), "SOLVER.IMS_PER_BATCH ({}) must be divisible by the number "
```

bernhardschaefer:
I realize now that the initial description I gave in the PR is incorrect, sorry for that.
This is not about how str.format() ignores the first string when they are separated, but rather that assert only considers the first string.

bernhardschaefer:
I also addressed the flake8 issue that you mentioned.

bernhardschaefer:
@fmassa I added some clarification and addressed the flake8 issue that you mentioned, feel free to review again.

PR #673, opened 2019-04-15
APIs moved to torch.hub

thomasshupeng:
APIs were moved from torch.utils.model_zoo to torch.hub 
As results of pytorch change - [add/move a few apis in torch.hub](https://github.com/pytorch/pytorch/pull/18758)


fmassa:
Subsumed by https://github.com/facebookresearch/maskrcnn-benchmark/pull/698

Issue #672, opened 2019-04-14
The detection accuracy of the R-50-FPN Faster R-CNN is lower than your report, confusing...


ChenJoya:
## ❓ Questions and Help
Hi @fmassa , thanks for your elegant implementation.
But it is confusing that the detection AP is only 32.8 when I re-train R-50-FPN Faster R-CNN, which should be 36.8 in your report:https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/MODEL_ZOO.md
```
2019-04-14 07:12:12,977 maskrcnn_benchmark.inference INFO: Start evaluation on coco_2017_val dataset(5000 images).
2019-04-14 07:15:06,105 maskrcnn_benchmark.inference INFO: Total run time: 0:02:53.127008 (0.06925080318450928 s / img per device, on 2 devices)
2019-04-14 07:15:06,105 maskrcnn_benchmark.inference INFO: Model inference time: 0:02:32.530358 (0.061012143325805665 s / img per device, on 2 devices)
2019-04-14 07:15:07,906 maskrcnn_benchmark.inference INFO: Preparing results for COCO format
2019-04-14 07:15:07,906 maskrcnn_benchmark.inference INFO: Preparing bbox results
2019-04-14 07:15:09,584 maskrcnn_benchmark.inference INFO: Evaluating predictions
2019-04-14 07:16:17,912 maskrcnn_benchmark.inference INFO: OrderedDict([('bbox', OrderedDict([('AP', 0.3275950734831557), ('AP50', 0.5054028517973591), ('AP75', 0.36449119818971715), ('APs', 0.1492328236066365), ('APm', 0.3439931485309256), ('APl', 0.48224050452315087)]))])

```
The config is not changed, but I only have 2 V100 GPUS, therefore 8 images are on each device.
Other information:
```
OS: Ubuntu 18.04.1 LTS
GCC version: (GCC) 5.5.0
CMake version: version 3.10.2

Python version: 3.7
Is CUDA available: Yes
CUDA runtime version: 9.0.176
GPU models and configuration:
GPU 0: Tesla P100-PCIE-16GB
GPU 1: Tesla P100-PCIE-16GB
GPU 2: Tesla P100-PCIE-16GB
GPU 3: Tesla V100-PCIE-16GB
GPU 4: Tesla V100-PCIE-16GB

Nvidia driver version: 418.43
cuDNN version: Probably one of the following:
[...]

PR #676, opened 2019-04-16
proposals from RPN per image during training

ChenJoya:
Hi @fmassa , the function ```select_over_all_levels``` is modified for single image during training.
The original issue: https://github.com/facebookresearch/maskrcnn-benchmark/issues/672
Thank you ^ ^

### DIFF
--- a/maskrcnn_benchmark/modeling/rpn/inference.py
+++ b/maskrcnn_benchmark/modeling/rpn/inference.py
@@ -158,7 +158,7 @@ class RPNPostProcessor(torch.nn.Module):
         # and not per batch
         if self.training:
             objectness = torch.cat(
-               [boxlist.get_field("objectness") for boxlist in boxlists], dim=0
+                [boxlist.get_field("objectness") for boxlist in boxlists], dim=0
             )
             box_sizes = [len(boxlist) for boxlist in boxlists]
             post_nms_top_n = min(self.fpn_post_nms_top_n, len(objectness))
--- a/maskrcnn_benchmark/modeling/rpn/inference.py
+++ b/maskrcnn_benchmark/modeling/rpn/inference.py
@@ -151,16 +151,23 @@ class RPNPostProcessor(torch.nn.Module):
 
     def select_over_all_levels(self, boxlists):
         num_images = len(boxlists)
+        # different behavior during training and during testing:
+        # during training, post_nms_top_n is over *all* the proposals combined, while
+        # during testing, it is over the proposals for each image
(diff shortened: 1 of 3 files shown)


ChenJoya:
Hi @fmassa , you are right, and the change of code needs to be check more.
Therefore I only rectify the README.md, and show how to set ```FPN_POST_NMS_TOP_N_TRAIN```. 
Thanks for your advise!

rodrigoberriel:
@fmassa Would it be okay if we add a flag (e.g., `_C.MODEL.RPN.AS_IN_DETECTRON` -- hopefully a better name) with default `True` and add a section (e.g., "Known Issues" or "Known Bugs") to the README describing the default behavior, known side-effects (the one described in this PR) and the workarounds (1. do as recommended in this PR if the user cannot afford to re-train or 2. train in the correct way changing the flag)? Or maybe just a section to the `TROUBLESHOOTING.md`? I feel like we should provide a "switch" to avoid training like this, since training in the correct way apparently leads to the "same" results; then, it would be up to the user.

What do you think, @ChenJoya ?

maskrcnn_benchmark/modeling/rpn/inference.py
```diff
@@ -158,7 +158,7 @@ def select_over_all_levels(self, boxlists):
         # and not per batch
         if self.training:
             objectness = torch.cat(
-                [boxlist.get_field("objectness") for boxlist in boxlists], dim=0
+               [boxlist.get_field("objectness") for boxlist in boxlists], dim=0
```
fmassa: just removing this extra space change and this should be good to merge!

fmassa:
@rodrigoberriel the documentation, after @ChenJoya improvements, already makes it fairly clear about this behavior.

Adding one more config argument would be an option to accommodate both behaviors, but it will just make the code more complicated.

rodrigoberriel:
@fmassa fair enough about the documentation

Would you be willing to review a PR trying to accommodate both behaviors? I think it would be a 5~10 lines change and we could convert this TODO into a NOTE :)

https://github.com/facebookresearch/maskrcnn-benchmark/blob/1714b7c20d919f8c536a577e15ba3a03ce733cde/maskrcnn_benchmark/modeling/rpn/inference.py#L154-L159

ChenJoya:
@fmassa  Thanks for your review. The extra space change has been removed :)

fmassa:
@rodrigoberriel sure, why not adding an option (defaulting to the current behavior) that switches between behaviors. PRs welcome! :-)

PR #680, opened 2019-04-16
add color jitter augmentation

CoinCheung:
Inspired by [this](https://arxiv.org/abs/1902.04103) paper, maybe the augmentation of adjusting color, brightness, contrast and saturation will make the model get trained more stably.


maskrcnn_benchmark/config/defaults.py
```diff
@@ -54,6 +54,12 @@
 # Convert image to BGR format (for Caffe2 models), in range 0-255
 _C.INPUT.TO_BGR255 = True
 
+# Image ColorJitter
+_C.INPUT.BRIGHTNETSS = 0.0
+_C.INPUT.CONTRAST = 0.0
+_C.INPUT.SITURATION = 0.0
```
fmassa: typo: it should be SATURATION

maskrcnn_benchmark/config/defaults.py
```diff
@@ -54,6 +54,12 @@
 # Convert image to BGR format (for Caffe2 models), in range 0-255
 _C.INPUT.TO_BGR255 = True
 
+# Image ColorJitter
+_C.INPUT.BRIGHTNETSS = 0.0
```
fmassa: typo: BRIGHTNESS

CoinCheung: Hi, 

I have correct the spelling, please tell me if there needs further modifications. 

As for the performance, I actually observed no obvious improvements on our own datasets with `brightness=0.4, contrast=0.4, saturation=0.4, hue=0.3`. The random seed is not manually assigned and the performance is on par with the non-color-aug version.

CoinCheung: By the way, do you think we should add a `random_seed` field in the configuration file for convenience of assigning the seed for random operations?

CoinCheung:
Hi, 

Is there further modification requirements, or is this pull request not really necessary ?

maskrcnn_benchmark/config/defaults.py
fmassa: Having a seed would be useful I think, even though not everything would be deterministic.

xiaohai12:
Hi, is this the configuration for data augmentation? My model seems to be overfitting and I would like to do some augmentation.

PR #697, opened 2019-04-20
add documentation for finetuning cityscapes

zimenglan-sysu-512:
please see `configs/cityscapes/README.md` for more details.


shinya7y:
@zimenglan-sysu-512 Thank you!
Are `863` and `1024, 1024` in `MIN_SIZE_TRAIN` typo?
https://github.com/facebookresearch/maskrcnn-benchmark/pull/697/files#diff-fc74046a7ed52a6392cffd3350a2939cR201

zimenglan-sysu-512:
hi @shinya7y 
i make a mistake, it should be 864. sorry for that. i add 1024 in `MIN_SIZE_TRAIN`, because in the paper `We train with image scale (shorter side) randomly sampled from [800, 1024]`.
i will send a new PR to fix it.

shinya7y:
@zimenglan-sysu-512 Thanks for fixing it.
Though it's unclear whether we should add 1024, the difference will not be a problem since the authors' exact settings are not available.

SkeletonOne:
@zimenglan-sysu-512 Could you tell me How to use the model surgery code in README?

PR #698, opened 2019-04-20
fix imports for latest pytorch-nightly

qianyizhang:


qianyizhang:
also with new imports from apex, python2 is no longer supported, you may want to address this in README or make apex lib optional

fmassa:
@qianyizhang oh, nice catch about apex being python3-only.
Ideally, we would make APEX optional, so that we can still support Python 2.

It would be great to have a PR fixing that.

zskadazhang:
hi, is there any way to avoid using apex, my cuda version is 9.0 so I can not install apex. I want to train with a single core cpu. look forward to your reply @ @fmassa

fmassa:
@zskadazhang for now, you'll need to checkout to a version of `maskrcnn-benchmark` prior to the merge of APEX.
This is the last commit before APEX was introduced

```
git checkout bf0437923c8ed16623fc9dc0ba3f688d2d9e8cee
```

The best solution would be to make APEX optional, and I'm willing to accept a PR that makes it optional

PR #708, opened 2019-04-22
on-line hard mining

CoinCheung:
This is a demo of ohem. I have tested on our own dataset, and the map/mar get boosted roughly 1 point.

The original  paper use nms to eliminate overlapping rois, but the original paper use ss to generate proposals, while this repository relies on rpn to generate proposals. So I did not implement associated part.

If this modification has potential risks or errors, please point them out and I will fix it.

### DIFF
--- a/maskrcnn_benchmark/modeling/roi_heads/box_head/loss.py
+++ b/maskrcnn_benchmark/modeling/roi_heads/box_head/loss.py
@@ -7,7 +7,8 @@ from maskrcnn_benchmark.modeling.box_coder import BoxCoder
 from maskrcnn_benchmark.modeling.matcher import Matcher
 from maskrcnn_benchmark.structures.boxlist_ops import boxlist_iou
 from maskrcnn_benchmark.modeling.balanced_positive_negative_sampler import (
-    BalancedPositiveNegativeSampler
+    BalancedPositiveNegativeSampler,
+    OhemPositiveNegativeSampler
 )
 from maskrcnn_benchmark.modeling.utils import cat
 
@@ -19,21 +20,23 @@ class FastRCNNLossComputation(object):
     """
 
     def __init__(
-        self, 
-        proposal_matcher, 
-        fg_bg_sampler, 
-        box_coder, 
+        self,
+        proposal_matcher,
+        fg_bg_sampler,
+        box_coder,
+        batch_size_per_image,
         cls_agnostic_bbox_reg=False
     ):
         """
         Arguments:
             proposal_matcher (Matcher)
-            fg_bg_sampler (BalancedPositiveNegativeSampler)
+            fg_bg_sampler (BalancedPositiveNegativeSampler, or OhemPositiveNegativeSampler)
             box_coder (BoxCoder)
         """
         self.proposal_matcher = proposal_matcher
         self.fg_bg_sampler = fg_bg_sampler
         self.box_coder = box_coder
+        self.batch_size_per_image = batch_size_per_image
         self.cls_agnostic_bbox_reg = cls_agnostic_bbox_reg
 
     def match_targets_to_proposals(self, proposal, target):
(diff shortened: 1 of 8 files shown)


maskrcnn_benchmark/modeling/roi_heads/box_head/loss.py
```diff
@@ -105,16 +108,79 @@ def subsample(self, proposals, targets):
...
+                [0, 1, 2, 3], device=device)
+
+        box_loss = smooth_l1_loss(
+            box_regression[sampled_pos_inds_subset[:, None], map_inds],
+            regression_targets[sampled_pos_inds_subset],
+            reduction='none',
+            beta=1,
+        ).sum(dim=1, keepdim=True)
+        ohem_loss = classification_loss
+        ohem_loss[sampled_pos_inds_subset[:, None]] += box_loss
```
WeihongM: This operation maybe is wrong, which should be modifed into
`ohem_loss[sampled_pos_inds_subset[:, None]] = ohem_loss[sampled_pos_inds_subset[:, None]] + box_loss`

CoinCheung: Hi, 
Thanks for reviewing. Would you please explain why this is wrong? I just do not understand the difference.

WeihongM: I am not sure for this, I have tried your code above, but I find it fail to add the box_loss to the correspond classification loss at ohem.  Maybe you can debug whether it's the same case.
Actually I changed the code
`        ohem_loss = classification_loss.clone()
        ohem_loss[sampled_pos_inds_subset[:, None]] = ohem_loss[sampled_pos_inds_subset[:, None]] + box_loss
`

maskrcnn_benchmark/structures/segmentation_mask.py
```diff
@@ -361,8 +361,9 @@ def __init__(self, polygons, size):
         self.polygons = []
         for p in polygons:
             p = PolygonInstance(p, size)
-            if len(p) > 0:
-                self.polygons.append(p)
+            #  if len(p) > 0:
```
fmassa: Why do we have this change?

maskrcnn_benchmark/modeling/roi_heads/box_head/loss.py
fmassa: I would the very least not use an in-place operation, as it might break backpropagation depending on the operation.

You could replace it with something like what @WeihongM mentioned

maskrcnn_benchmark/structures/segmentation_mask.py
CoinCheung: Hi, 

I believe this modification anymore is no longer needed any more. I modified this because our json files have empty `segmentation` field in the `annotations` part with will break the logic here.

CoinCheung: I have re-committed this pr, and new link is [here](https://github.com/facebookresearch/maskrcnn-benchmark/pull/845). If you please, we could continue communicating in that page :)

maskrcnn_benchmark/modeling/roi_heads/box_head/loss.py
CoinCheung: Thanks for reviewing this. I have modified as you suggested.

chucklu:
@CoinCheung   You can get the delete branch in fork back  through https://github.com/betaflight/betaflight/pull/8112#issuecomment-492081397

LouieYang:
@CoinCheung Why the ohem_loss.size(0) compared with self.batch_size_per_image? It seems like ohem_loss includes all proposals from all batch images and maybe it should be torch.split to image-wise and then use top_k to select?

WeihongM:
@CoinCheung    same question with Louie Yang, think this implementation have some problem.

PR #710, opened 2019-04-22
Quickfix for outdated param names

Iwontbecreative:
Noticed the names in the comment do not correspond to any actual params, fixing to refer to an actual one.


Iwontbecreative:
I just signed the CLA btw

PR #739, opened 2019-05-01
Windows10 installation fixes ver2

Idolized22:


fmassa:
There seems to be a rebase problem with this PR (there are 32 commits here, and 52 files changed), can you maybe send a new PR with the fixes? Thanks!

Issue #169, opened 2018-11-18
How to add negative samples for my training set?


Igal20:
## ❓ Questions and Help
Hello, 
My data have only two classes first is background = 0 and object = 1.
While training I it necessary for me to present to the net negative samples i.e images without an object, just background. In this case I don't have segmentation contour. How do I add those images to the training? 

I use COCO-style annotations for the images, saved in json format.
Thanks in advance.

wjp0408:
I think this[https://github.com/facebookresearch/maskrcnn-benchmark/issues/144](url) can help you.

Igal20:
@wjp0408  Not so much, issue 144 handle with negative samples  simply by filtering them and not including in training at all.
I do not want to exclude them but use them while training.

fmassa:
Hi

One image should contain two things: a positive element and a negative element.
This is always the case for detection, except if your bounding boxes are of the full size of the image, in which case you might be looking for object classification, and not object detection?

Igal20:
@fmassa  Thanks for the answer. 
If I understand right "One image should contain two things: a positive element and a negative element."
is true for evaluating the net not for training.

I'll elaborate a little bit more on what I aim to do.
For example I want to train Mask-rcnn to detect and segment giraffes. So I provide 1000 sample of images and masks of giraffes. But I have also 10 image that look like the animal [for example blanket with giraffes pattern] and I want to include them in my training process but obviously masks of those images will be blank.
Thus I want to be able to add images without segmentation mask.  
 Thanks

fmassa:
I see, thanks for the explanation!

It is possible to support your use-case, but you'll need to adapt a few things in the code for that.
The first issue you'll see is this one: https://github.com/facebookresearch/maskrcnn-benchmark/issues/31
You can make [the Matcher return a tensor of size N](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/modeling/matcher.py#L56), but you'll afterwards probably face a few other issues down the road that will need to be addressed.

Let me know if you get stuck in a particular problem.  It might be interesting to see, once the code is working, what it would take to make it work, so that we might potentially send a PR to add support for this in the codebase.

Igal20:
Thanks I'll check it.

Iwontbecreative:
I have the same objective of being able to train on negative examples (images where there is nothing to detect). 

[...]

PR #777, opened 2019-05-11
fix the bug when initializing BinaryMaskList with an empty list

krumo:
As discussed in this [commit](https://github.com/facebookresearch/maskrcnn-benchmark/commit/b4d546577850b156939067794f155526e262ee61), [Line55](https://github.com/facebookresearch/maskrcnn-benchmark/blob/1127bdd368613f320f7b113320e62994c0baa216/maskrcnn_benchmark/structures/segmentation_mask.py#L55) for initializing a BinaryMaskList assumes the list passed in `BinaryMaskList` contains at least one element, which doesn't always hold in practice. 

My main reasons for adding this empty list handling code are as follows:
1. Support more use cases:
* When we are conducting validation, we may need to check the false positive results on both images contain target objects and images don't contain target objects. A support for accepting empty binary mask during validation is necessary.
* As discussed in #169 , some people may need training on negative examples for their custom dataset. A support for accepting empty binary mask during training is required.
2. Consistency of coding style
* In `segmentation_mask`, masks may be described as a `BinaryMaskList` or a `PolygonList`. In the implementation of `PolygonList`, [Line341](https://github.com/facebookresearch/maskrcnn-benchmark/blob/1127bdd368613f320f7b113320e62994c0baa216/maskrcnn_benchmark/structures/segmentation_mask.py#L341) could successfully deal with empty mask. Thus, a support for dealing with empty mask in `BinaryMaskList` is also necessary.

I make a simple modification for handling empty mask. The test for segmentation_mask has been passed successfully. Any comments are welcome!

### DIFF
--- a/maskrcnn_benchmark/structures/segmentation_mask.py
+++ b/maskrcnn_benchmark/structures/segmentation_mask.py
@@ -52,7 +52,9 @@ class BinaryMaskList(object):
             # The raw data representation is passed as argument
             masks = masks.clone()
         elif isinstance(masks, (list, tuple)):
-            if isinstance(masks[0], torch.Tensor):
+            if len(masks) == 0:
+                masks = torch.zeros([size[1], size[0]]).clone()
+            elif isinstance(masks[0], torch.Tensor):
                 masks = torch.stack(masks, dim=2).clone()
             elif isinstance(masks[0], dict) and "counts" in masks[0]:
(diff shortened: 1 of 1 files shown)


fmassa:
Subsumed by https://github.com/facebookresearch/maskrcnn-benchmark/pull/780

Thanks @krumo for the original PR!

Issue #348, opened 2019-01-16
How to record mAP for every epoch on the validation dataset


nprasad2021:
## ❓ Questions and Help
Hi,

I have two questions relating to storing training/val/test accuracies/losses - 

1. In the log file, where I can find mAP values for training, validation, and testing datasets?
2. I have implemented the workflow in issue #171 , where I call the inference function in engine.trainer.do_train. Where should I modify the code (in the "inference" function or in the "coco_evaluation" file to save mAP values and losses at every epoch?

fmassa:
Hi,

1 - We don't compute the mAP during training. You could compute it for the training set, but it would take quite some time.
2 - You can modify this function, to store the mAP / losses in the format that you want https://github.com/facebookresearch/maskrcnn-benchmark/blob/d28845e112de36781b2b5f7217a34b2b62de8d2f/maskrcnn_benchmark/data/datasets/evaluation/coco/coco_eval.py#L63

Let me know if you have further questions.

nprasad2021:
Thanks! I've added the following function "val" to maskrcnn_benchmark.engine.trainer - >
`def val(cfg, model, distributed=False):
    if distributed:
        model = model.module
    torch.cuda.empty_cache()  # TODO check if it helps
    iou_types = ("bbox",)
    if cfg.MODEL.MASK_ON:
        iou_types = iou_types + ("segm",)
    output_folders = [None] * (len(cfg.DATASETS.TEST) + len(cfg.DATASETS.TRAIN))
    dataset_names = cfg.DATASETS.TEST + cfg.DATASETS.TRAIN
    if cfg.OUTPUT_DIR:
        for idx, dataset_name in enumerate(dataset_names):
            print(dataset_name)
            output_folder = os.path.join(cfg.OUTPUT_DIR, "inference", dataset_name)
            mkdir(output_folder)
            output_folders[idx] = output_folder
    data_loaders_val = make_data_loader(cfg, is_train=False, is_distributed=distributed)
    output_tuple = {}
    for output_folder, dataset_name, data_loader_val in zip(output_folders, dataset_names, data_loaders_val):
        (dataset_name)
        result = inference(
            model,
            data_loader_val,
            dataset_name=dataset_name,
            iou_types=iou_types,
            box_only=cfg.MODEL.RPN_ONLY,
            device=cfg.MODEL.DEVICE,
            expected_results=cfg.TEST.EXPECTED_RESULTS,
[...]

Issue #171, opened 2018-11-18
How do I perform validation?


Igal20:
## ❓ Questions and Help
Hello, 
Once in a while I would like to perform test on validation set, what is right way to do so ?
THanks

fmassa:
Hi,

You could add a condition in [here](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/engine/trainer.py#L105) so that it calls the [inference function](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/tools/train_net.py#L95) to validate on a dataset.
You might need to pass a few extra arguments around, but it should be fairly straightforward.

But given that we are doing training on multiple GPUs, if you want to perform testing on multiple GPUs as well you might need to unwrap the `model` from the `DistributedDataParallel`, add a `synchronize` call in the end of the inference and wrap the model back in `DistributedDataParallel` just before the training starts again.

Let me know if you have further questions

JoyHuYY1412:
> unwrap the `model` from the `DistributedDataParallel`, add a `synchronize` call

Sorry but how to unwrap the model?

Like what should I change in this code:
`if iteration % checkpoint_period == 0 or iteration == max_iter:
         model.eval()
          iou_types = ("bbox",)
            if cfg.MODEL.MASK_ON:
                iou_types = iou_types + ("segm",)
                    inference(
                        model,
                        val_data_loader,
                        iou_types=iou_types,
                        box_only=cfg.MODEL.RPN_ONLY,
                        device=cfg.MODEL.DEVICE,
                        expected_results=cfg.TEST.EXPECTED_RESULTS,
                        expected_results_sigma_tol=cfg.TEST.EXPECTED_RESULTS_SIGMA_TOL,
                        verbose=False
                    )

                model.train()  # reset training flag
            synchronize()
`

fmassa:
@JoyHuYY1412 yes, this makes sense, but I won't guarantee that it will work 100% in all cases.

JoyHuYY1412:
> @JoyHuYY1412 yes, this makes sense, but I won't guarantee that it will work 100% in all cases.

It does not work with multigpu. I don't how to tell it just use one GPU in validation

PR #785, opened 2019-05-15
Validation datasets support during training

osanwe:
It seems that the framework does not have capabilities to use validation datasets during training for checking convergence (it attaches them into a training dataset), and it is a requested feature (https://github.com/facebookresearch/maskrcnn-benchmark/issues/171 and https://github.com/facebookresearch/maskrcnn-benchmark/issues/348).

In both mentioned issues not clean solutions are proposed because of neediness of additional boilerplate code. So the PR contains integration of the validation into training process with configs support.

Also this feature will be useful for visualization after adding TensorBoard support (https://github.com/facebookresearch/maskrcnn-benchmark/pull/163).

The usage example presented in this [config file](https://github.com/osanwe/maskrcnn-benchmark/blob/with_val/configs/e2e_mask_rcnn_R_50_FPN_1x.yaml).

**Upd.:** The latest version of PR is https://github.com/facebookresearch/maskrcnn-benchmark/pull/828

### DIFF
--- a/configs/e2e_mask_rcnn_R_50_FPN_1x.yaml
+++ b/configs/e2e_mask_rcnn_R_50_FPN_1x.yaml
@@ -30,8 +30,9 @@ MODEL:
     SHARE_BOX_FEATURE_EXTRACTOR: False
   MASK_ON: True
 DATASETS:
-  TRAIN: ("coco_lip_2019_train",)
-  VAL: ("coco_lip_2019_val",)
+  TRAIN: ("coco_2014_train",)
+  VAL: ("coco_2014_valminusminival",)
+  TEST: ("coco_2014_minival",)
 DATALOADER:
   SIZE_DIVISIBILITY: 32
 SOLVER:
--- a/configs/e2e_mask_rcnn_R_50_FPN_1x.yaml
+++ b/configs/e2e_mask_rcnn_R_50_FPN_1x.yaml
@@ -30,8 +30,8 @@ MODEL:
     SHARE_BOX_FEATURE_EXTRACTOR: False
   MASK_ON: True
 DATASETS:
-  TRAIN: ("coco_2014_train", "coco_2014_valminusminival")
-  TEST: ("coco_2014_minival",)
+  TRAIN: ("coco_lip_2019_train",)
+  VAL: ("coco_lip_2019_val",)
 DATALOADER:
   SIZE_DIVISIBILITY: 32
 SOLVER:
--- a/maskrcnn_benchmark/data/build.py
+++ b/maskrcnn_benchmark/data/build.py
@@ -23,7 +23,7 @@ def build_dataset(dataset_list, transforms, dataset_catalog, mode=DatasetMode.TR
         transforms (callable): transforms to apply to each (image, target) sample
         dataset_catalog (DatasetCatalog): contains the information on how to
             construct a dataset.
-        is_train (bool): whether to setup the dataset for training or testing
+        mode (DatasetMode): whether to setup the dataset for training, validation, or testing
     """
     if not isinstance(dataset_list, (list, tuple)):
         raise RuntimeError(
@@ -37,9 +37,9 @@ def build_dataset(dataset_list, transforms, dataset_catalog, mode=DatasetMode.TR
         # for COCODataset, we want to remove images without annotations
         # during training
         if data["factory"] == "COCODataset":
-            args["remove_images_without_annotations"] = is_train
+            args["remove_images_without_annotations"] = mode == DatasetMode.TRAIN
         if data["factory"] == "PascalVOCDataset":
-            args["use_difficult"] = not is_train
+            args["use_difficult"] = mode != DatasetMode.TRAIN
         args["transforms"] = transforms
         # make dataset from factory
         dataset = factory(**args)
@@ -163,8 +163,8 @@ def make_data_loader(cfg, mode=DatasetMode.TRAIN, is_distributed=False, start_it
     )
     DatasetCatalog = paths_catalog.DatasetCatalog
 
-    transforms = build_transforms(cfg, is_train)
-    datasets = build_dataset(dataset_list, transforms, DatasetCatalog, is_train)
+    transforms = build_transforms(cfg, mode)
+    datasets = build_dataset(dataset_list, transforms, DatasetCatalog, mode)
 
     data_loaders = []
     for dataset in datasets:
--- a/maskrcnn_benchmark/data/build.py
+++ b/maskrcnn_benchmark/data/build.py
@@ -2,7 +2,6 @@
 import bisect
 import copy
 import logging
-from enum import Enum
 
 import torch.utils.data
 from maskrcnn_benchmark.utils.comm import get_world_size
@@ -10,15 +9,10 @@ from maskrcnn_benchmark.utils.imports import import_file
 
 from . import datasets as D
 from . import samplers
+from .dataset_mode import DatasetMode
 
 from .collate_batch import BatchCollator
 from .transforms import build_transforms
-
-
-class DatasetMode(Enum):
-    TRAIN = 1
-    VALID = 2
-    TEST = 3
 
 
 def build_dataset(dataset_list, transforms, dataset_catalog, mode=DatasetMode.TRAIN):
--- a/maskrcnn_benchmark/data/build.py
+++ b/maskrcnn_benchmark/data/build.py
@@ -118,7 +118,7 @@ def make_data_loader(cfg, mode=DatasetMode.TRAIN, is_distributed=False, start_it
         num_iters = cfg.SOLVER.MAX_ITER
         dataset_list = cfg.DATASETS.TRAIN
     elif mode == DatasetMode.VALID:
-        images_per_batch = cfg.SOLVER.IMS_PER_BATCH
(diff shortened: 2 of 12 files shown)


zimenglan-sysu-512:
hi @osanwe 
can u give some results of augmentations (flip, scales flip+scales)?

maskrcnn_benchmark/engine/trainer.py
```diff
@@ -107,6 +108,37 @@ def do_train(
...
+                            "eta: {eta}",
+                            "iter: {iter}",
+                            "{meters}",
+                            "lr: {lr:.6f}",
+                            "max mem: {memory:.0f}",
+                        ]
+                    ).format(
+                        eta=eta_string,
+                        iter=iteration,
+                        meters=str(meters),
```
droseger: meters needs to be meters_val here, otherwise the training metrics are displayed

osanwe: Oh... Yes. Fixed.

osanwe:
@fmassa ,
Sorry for ping.
As I understood from previous activity, you are a maintainer of the repo. Is this pull request useful or it can be closed?
Maybe some additional information or tests should be provided from my side. The code was tested on 1 and 8 GPUs.

configs/e2e_mask_rcnn_R_50_FPN_1x.yaml
```diff
@@ -30,7 +30,8 @@ MODEL:
     SHARE_BOX_FEATURE_EXTRACTOR: False
   MASK_ON: True
 DATASETS:
-  TRAIN: ("coco_2014_train", "coco_2014_valminusminival")
+  TRAIN: ("coco_2014_train",)
```
fmassa: Can you revert this change? All the models have been trained using the new `coco_2017train` dataset, which corresponds to `coco_2014_train` + `coco_2014_valminusminival`. If you want to evaluate at every N iterations, you could do it on the `coco_2014_minival`?

maskrcnn_benchmark/data/build.py
```diff
@@ -36,16 +37,16 @@ def build_dataset(dataset_list, transforms, dataset_catalog, is_train=True):
...
-            args["use_difficult"] = not is_train
+            args["use_difficult"] = mode == DatasetMode.TEST
         args["transforms"] = transforms
         # make dataset from factory
         dataset = factory(**args)
         datasets.append(dataset)
 
     # for testing, return a list of datasets
-    if not is_train:
+    if mode != DatasetMode.TEST:
```
fmassa: Even though not really the best thing to do, I believe in most cases we simply evaluate on the test dataset after N iterations, so I think that we can remove the `VAL` part altogether.

maskrcnn_benchmark/engine/trainer.py
```diff
@@ -107,6 +108,37 @@ def do_train(
...
+                synchronize()
+                with torch.no_grad():
+                    for idx_val, (images_val, targets_val, _) in enumerate(data_loader_val):
+                        images_val = images_val.to(device)
+                        targets_val = [target.to(device) for target in targets_val]
+                        loss_dict = model(images_val, targets_val)
+                        losses = sum(loss for loss in loss_dict.values())
+                        loss_dict_reduced = reduce_loss_dict(loss_dict)
+                        losses_reduced = sum(loss for loss in loss_dict_reduced.values())
+                        meters_val.update(loss=losses_reduced, **loss_dict_reduced)
```
fmassa: If I understand it correctly, you only evaluate the loss here, while a metric which is generally more useful is to report the mAP as we do for testing.

fmassa (changes requested):
Thanks for the PR!

I think that having a separate VAL set might not be necessary.

Also, you only log the losses during validation. While more complicated, in the new release of torchvision I added a functionality to progressively compute the mAP during evaluation, and as an example I compute evaluation at the end of every epoch, see https://github.com/pytorch/vision/blob/master/references/detection/coco_eval.py

Maybe something like that could be used here instead?

YradenRavid:
Hi, I'm working on the same thing write now, and I wonder why is it possible to avoid model.eval() command before the validation starts?

osanwe:
@fmassa,
Yes, it seems the additional datasets field is redundant. So I created another PR (https://github.com/facebookresearch/maskrcnn-benchmark/pull/828) where the same datasets are used for intermediate and final evaluations with suggested AP calculations.

@YradenRavid,
The `train` and `eval` methods change the inner state of the module (https://github.com/pytorch/pytorch/blob/master/torch/nn/modules/module.py#L987-L1011) and can impact to the model behavior.
For validation in this PR loss values are needed, and they can be gotten only in training mode (https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/modeling/detector/generalized_rcnn.py#L59).

qihao96:
@osanwe I'm working on validation set average loss curve for Tensor Board,  and almost done.
> Also this feature will be useful for visualization after adding TensorBoard support (#163).

YradenRavid:
@osanwe to my understanding, model.eval() change the model to evaluation mode (dropout, BN etc), why is it "good practice" to not use it here before running evaluation loss?

osanwe:
@YradenRavid,
It seems `FrozenBatchNorm2d` class is created by @fmassa, not by me.
As I wrote previously, `eval()` and `train()` methods only change the `training` boolean flag in `nn.Module`. And yes, this flag can impact on the `batch_norm` behavior and some other layers. Also it impacts on the RCNN behavior here (as mentioned previously), so you cannot get loss values if `training` flag is equal to `False` (i.e. you use `eval` mehod).

maskrcnn_benchmark/engine/trainer.py
qihao96: This line records batch'loss of val-set using current train iteration model, right?
So, if the purpose is to check our model is over fitting or not, we need to calculate the average loss of val-set using current train iteration model. And use this average loss to decide early stop.

alono88:
> @YradenRavid,
> It seems `FrozenBatchNorm2d` class is created by @fmassa, not by me.
> As I wrote previously, `eval()` and `train()` methods only change the `training` boolean flag in `nn.Module`. And yes, this flag can impact on the `batch_norm` behavior and some other layers. Also it impacts on the RCNN behavior here (as mentioned previously), so you cannot get loss values if `training` flag is equal to `False` (i.e. you use `eval()` mehod).

By inspecting the implementation of the FrozenBatchNorm, it seems all parameters are indeed frozen and the behavior of the layer does not change regardless of `.eval()` or `.train()`.

maskrcnn_benchmark/engine/trainer.py
osanwe: Yes, global average loss is needed and `MetricLogger` class calculates it:
https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/utils/metric_logger.py#L35-L37

osanwe:
> behavior of the layer does not change regardless of `.eval()` or `.train()`

The behavior is changed in `GeneralizedRCNN` `forward` method:
https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/maskrcnn_benchmark/utils/metric_logger.py#L35-L37

configs/e2e_mask_rcnn_R_50_FPN_1x.yaml
osanwe: I've reverted it and created another config file where the number of iterations for validation specified:
https://github.com/facebookresearch/maskrcnn-benchmark/pull/828/files#diff-4dd26a63ac00a49aeb10985800d7f21c

maskrcnn_benchmark/data/build.py
osanwe: I added another boolean flag instead for controlling the way of data loader creating:
https://github.com/facebookresearch/maskrcnn-benchmark/pull/828/files#diff-48c338613bdbf422235cdb2ef17201f7R77

maskrcnn_benchmark/engine/trainer.py
osanwe: I added inference metrics calculating in addition to loss calculation:
https://github.com/facebookresearch/maskrcnn-benchmark/pull/828/files#diff-29486803add8a1cde2a6e5b741434c7cR128

JoyHuYY1412:
> It seems that the framework does not have capabilities to use validation datasets during training for checking convergence (it attaches them into a training dataset), and it is a requested feature (#171 and #348).
> 
> In both mentioned issues not clean solutions are proposed because of neediness of additional boilerplate code. So the PR contains integration of the validation into training process with configs support.
> 
> Also this feature will be useful for visualization after adding TensorBoard support (#163).
> 
> The usage example presented in this [config file](https://github.com/osanwe/maskrcnn-benchmark/blob/with_val/configs/e2e_mask_rcnn_R_50_FPN_1x.yaml).
> 
> **Upd.:** The latest version of PR is #828

It seems this inference operation will increase the cost of memory? I encountered out of memory when gathering results from the gpus. I am so confused that the inference can be finished but the memory will be costed a lot when gathering results?

osanwe:
> It seems this inference operation will increase the cost of memory?

Yes, the validation step requires additional memory because of additional data loader etc. Please, refer to https://github.com/facebookresearch/maskrcnn-benchmark/pull/828.

lufficc:
Preparing pr description...

PR #798, opened 2019-05-20
keep the resize function the same in test time the same with training time

hcx1231:
The result may be affected  when   the implementation of resize function in inference time  and training time is inconsistent,so I changed the code in inference time to keep them same.


fmassa:
The only difference I see is that the images might be larger than 1333 pixels, but is this a problem?

hcx1231:
> The only difference I see is that the images might be larger than 1333 pixels, but is this a problem?

Yes,when the images are very large,for example in my own dataset,it may cause large error as it haven't been trained in such size,which took me several weeks to find out this problem.

fmassa:
Ok, fair enough.

PR #804, opened 2019-05-21
Multi-scale testing

fallingdust:
Implemented bbox augmentations(h-flip, multi-scale, multi-scale h-flip) in test time, just like Detectron.

There's a little hack in the data loading process, if you have better solution, please tell me.


zimenglan-sysu-512:
hi @fallingdust 
can u give some results of augmentations (flip, scales flip+scales)?

fallingdust:
hi @zimenglan-sysu-512 ,
I evaluated on **coco_2014_minival** with pretrained model downloaded from _https://download.pytorch.org/models/maskrcnn/e2e_faster_rcnn_R_101_FPN_1x.pth_, here is the bbox mAP:

Settings|mAP
--------------|--------------
w/o BBOX_AUG|39.1
w/ H_FLIP: True|39.7
w/ SCALES: (400, 600, 1000, 1200) MAX_SIZE: 2000|40.8
w/ H_FLIP: True SCALES: (400, 600, 1000, 1200) MAX_SIZE: 2000 SCALE_H_FLIP: True| 41.1

maskrcnn_benchmark/engine/bbox_aug.py
```diff
@@ -0,0 +1,118 @@
+import torch
+import torchvision.transforms as TT
+
+from maskrcnn_benchmark.config import cfg
```
fmassa: using the config globally in the library is something I've tried to avoid, but let's merge this as is and maybe modify this later on

shikunyu8:
Since you disabled the bbox filtering after box head, you will have 1000 boxes fed into the mask head. How could you fit 1000 masks into GPU memory?

shikunyu8:
I guess you test on faster rcnn, but the config file is mask rcnn.

youngwanLEE:
@fallingdust @shikunyu8 @fmassa 
This TTA is only for box detection?  Is it available for mask rcnn?

shikunyu8:
Hi, @youngwanLEE 
Yes. It's not available for mask rcnn.

PR #811, opened 2019-05-23
some protection code in bounding box operations

lsrock1:
1. In nms function, set max_proposals to min(max_proposals, len(keep)) to protect against over indexing.

2. In iou function, convert boxlist to xyxy mode explicitly to prevent unwarned malfunction.

### DIFF
--- a/maskrcnn_benchmark/structures/boxlist_ops.py
+++ b/maskrcnn_benchmark/structures/boxlist_ops.py
@@ -26,8 +26,6 @@ def boxlist_nms(boxlist, nms_thresh, max_proposals=-1, score_field="scores"):
     score = boxlist.get_field(score_field)
     keep = _box_nms(boxes, score, nms_thresh)
     if max_proposals > 0:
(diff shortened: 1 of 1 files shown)


maskrcnn_benchmark/structures/boxlist_ops.py
```diff
@@ -26,6 +26,8 @@ def boxlist_nms(boxlist, nms_thresh, max_proposals=-1, score_field="scores"):
     score = boxlist.get_field(score_field)
     keep = _box_nms(boxes, score, nms_thresh)
     if max_proposals > 0:
+        #protect against not enough keep elements
+        max_proposals = min(max_proposals, len(keep))
```
fmassa: this is not necessary. Indeed, even if `max_proposals` is larger than `keep`, the slicing will just take all elements.

lsrock1: thanks for your kindly response.
I removed unnecessary lines.

Issue #809, opened 2019-05-22
Why cannot I use pure bbox json file ?


CoinCheung:
## 🐛 Bug
Part of our self-created json annotation file goes like this: 
```
  "annotations": [
    {
      "id": 0,
      "category_id": 1,
      "bbox": [
        1072,
        1072,
        256,
        256
      ],
      "image_id": 0,
      "iscrowd": 0,
      "area": 65536,
      "segmentation": [
        []
      ]
    },
    {
      "id": 1,
      "category_id": 2,
      "bbox": [
        956,
        956,
        275,
        275
      ],
      "image_id": 0,
      "iscrowd": 0,
      "area": 75625,
      "segmentation": [
        []
      ]
    },
```
We want to use faster-rcnn to regress the bounding box of the object, and we do not have mask annotations. When I run the code with this json file, I got this error: 
```
 File "/mnt/lustre/ai-vision/home/yz885/miniconda3/envs/py37/lib/python3.7/site-packages/torch/utils/data/_utils/worker.py", line 99, in _worker_loop
    samples = collate_fn([dataset[i] for i in batch_indices])
  File "/mnt/lustre/ai-vision/home/yz885/miniconda3/envs/py37/lib/python3.7/site-packages/torch/utils/data/_utils/worker.py", line 99, in <listcomp>
    samples = collate_fn([dataset[i] for i in batch_indices])
  File "/mnt/lustre/ai-vision/home/yz885/projects/yd/maskrcnn-benchmark/maskrcnn_benchmark/data/datasets/coco.py", line 91, in __getitem__
    target = target.clip_to_image(remove_empty=True)
  File "/mnt/lustre/ai-vision/home/yz885/projects/yd/maskrcnn-benchmark/maskrcnn_benchmark/structures/bounding_box.py", line 223, in clip_to_image
    return self[keep]
  File "/mnt/lustre/ai-vision/home/yz885/projects/yd/maskrcnn-benchmark/maskrcnn_benchmark/structures/bounding_box.py", line 208, in __getitem__
    bbox.add_field(k, v[item])
  File "/mnt/lustre/ai-vision/home/yz885/projects/yd/maskrcnn-benchmark/maskrcnn_benchmark/structures/segmentation_mask.py", line 517, in __getitem__
    selected_instances = self.instances.__getitem__(item)
[...]

PR #814, opened 2019-05-24
modify senario where dataset has no mask provided

CoinCheung:
Hi,

As reported in [this](https://github.com/facebookresearch/maskrcnn-benchmark/issues/809) issue, when there is no `segmentation` field in the json file (which I believe will be quite common in practice, since most people would like to use this codebase in the bbox detection task), there will be a error of `index out of range` in the `boxlist` indexing operations. I feel maybe this modification would allow to get rid of this error by not adding the filed of `mask` to the `bboxlist` object.

I am not sure whether this modification is in consistent with  the designing spirit of this repo. If it is not proper to implement this modification, please let me know and I will delete this pr so that we can discuss another place where I feel maybe a little modification will make it better :)

### DIFF
--- a/maskrcnn_benchmark/data/datasets/coco.py
+++ b/maskrcnn_benchmark/data/datasets/coco.py
@@ -79,9 +79,8 @@ class COCODataset(torchvision.datasets.coco.CocoDetection):
         classes = torch.tensor(classes)
         target.add_field("labels", classes)
 
-        masks = [obj["segmentation"] for obj in anno if any(obj["segmentation"])]
-        if any(masks):
-            assert len(masks) == len(anno), "some bbox does not have associated mask!!"
+        if anno and "segmentation" in anno[0]:
+            masks = [obj["segmentation"] for obj in anno]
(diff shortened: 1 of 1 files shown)


maskrcnn_benchmark/data/datasets/coco.py
```diff
@@ -79,9 +79,11 @@ def __getitem__(self, idx):
         classes = torch.tensor(classes)
         target.add_field("labels", classes)
 
-        masks = [obj["segmentation"] for obj in anno]
-        masks = SegmentationMask(masks, img.size, mode='poly')
-        target.add_field("masks", masks)
+        masks = [obj["segmentation"] for obj in anno if any(obj["segmentation"])]
```
fmassa: maybe you could just add something similar to what is done for keypoints just below?
```
if anno and 'segmentation' in anno[0]:
    # rest of the code here
```

CoinCheung: Hi, 

Thanks for this suggestion, I have modified the code as the keypoint part. Please add more comments if further modification is required.

PR #835, opened 2019-05-29
Save full configuration in output dir

miguelvr:
In order to facilitate reproduction and deployment of models it is a good practice to save the exact training configuration together with the model weights.

In this PR the full config is dumped into the output directory with the name `config.yml`


tools/train_net.py
```diff
@@ -176,6 +176,11 @@ def main():
         logger.info(config_str)
     logger.info("Running with config:\n{}".format(cfg))
 
+    output_config_path = "{}/config.yml".format(cfg.OUTPUT_DIR)
```
fmassa: `os.path.join` is more robust wrt OS (Windows)

miguelvr: just resolved this, should be ready to merge!

Issue #805, opened 2019-05-21
No ground-truth boxes available for one of the images during training


miaoshuyu:
Traceback (most recent call last):
  File "tools/train_net.py", line 186, in <module>
    main()
  File "tools/train_net.py", line 179, in main
    model = train(cfg, args.local_rank, args.distributed)
  File "tools/train_net.py", line 85, in train
    arguments,
  File "/home/msy/project/Maskrcnn_benchmark/maskrcnn-benchmark/maskrcnn_benchmark/engine/trainer.py", line 68, in do_train
    loss_dict = model(images, targets)
  File "/home/msy/anaconda3/lib/python3.6/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/msy/anaconda3/lib/python3.6/site-packages/apex-0.1-py3.6-linux-x86_64.egg/apex/amp/_initialize.py", line 194, in new_fwd
    **applier(kwargs, input_caster))
  File "/home/msy/project/Maskrcnn_benchmark/maskrcnn-benchmark/maskrcnn_benchmark/modeling/detector/generalized_rcnn.py", line 50, in forward
    proposals, proposal_losses = self.rpn(images, features, targets)
  File "/home/msy/anaconda3/lib/python3.6/site-packages/torch/nn/modules/module.py", line 493, in __call__
    result = self.forward(*input, **kwargs)
  File "/home/msy/project/Maskrcnn_benchmark/maskrcnn-benchmark/maskrcnn_benchmark/modeling/rpn/rpn.py", line 159, in forward
    return self._forward_train(anchors, objectness, rpn_box_regression, targets)
  File "/home/msy/project/Maskrcnn_benchmark/maskrcnn-benchmark/maskrcnn_benchmark/modeling/rpn/rpn.py", line 178, in _forward_train
    anchors, objectness, rpn_box_regression, targets
  File "/home/msy/project/Maskrcnn_benchmark/maskrcnn-benchmark/maskrcnn_benchmark/modeling/rpn/loss.py", line 105, in __call__
    labels, regression_targets = self.prepare_targets(anchors, targets)
  File "/home/msy/project/Maskrcnn_benchmark/maskrcnn-benchmark/maskrcnn_benchmark/modeling/rpn/loss.py", line 61, in prepare_targets
    anchors_per_image, targets_per_image, self.copied_fields
[...]

Issue #724, opened 2019-04-29
ValueError: No ground-truth boxes available for one of the images during training


xrenaa:
## ❓ Questions and Help
I run the code successfully with my own datasets at the beginning. But after 40 iterations, and there is error. 
<img width="1679" alt="屏幕快照 2019-04-29 下午1 57 31" src="https://user-images.githubusercontent.com/31800248/56878557-4a3aaf80-6a87-11e9-9766-e79f2fc4027f.png">
I have checked that the training images all have their own annotations. 
I don't know how to deal with it.
Thank you!

travispan96:
I have the same issue, every image has the annotations.

xrenaa:
@travispan96  I find that there are errors in my annotations like that the bbox value may be negative. And there is a method that you can check which image is wrong:
in "maskrcnn-benchmark/maskrcnn_benchmark/engine/trainer.py"  line 57 
"for iteration, (images, targets, _) in enumerate(data_loader, start_iter):"
add a line of code:
 print("what pairs of images id that I am using", _ )
Then you will find the image id that was wrong!

travispan96:
> @travispan96 I find that there are errors in my annotations like that the bbox value may be negative. And there is a method that you can check which image is wrong:
> in "maskrcnn-benchmark/maskrcnn_benchmark/engine/trainer.py" line 57
> "for iteration, (images, targets, _) in enumerate(data_loader, start_iter):"
> add a line of code:
> print("what pairs of images id that I am using", _ )
> Then you will find the image id that was wrong!

I'll try it, thanks for your suggestion!

travispan96:
> @travispan96 I find that there are errors in my annotations like that the bbox value may be negative. And there is a method that you can check which image is wrong:
> in "maskrcnn-benchmark/maskrcnn_benchmark/engine/trainer.py" line 57
> "for iteration, (images, targets, _) in enumerate(data_loader, start_iter):"
> add a line of code:
> print("what pairs of images id that I am using", _ )
> Then you will find the image id that was wrong!

could you please tell me what the meaning is of the value in _, like"(1332, 2968)". How can I know the image id?

xrenaa:
> > @travispan96 I find that there are errors in my annotations like that the bbox value may be negative. And there is a method that you can check which image is wrong:
> > in "maskrcnn-benchmark/maskrcnn_benchmark/engine/trainer.py" line 57
[...]

PR #836, opened 2019-05-29
WorkAround for `ValueError: No ground-truth boxes available for one of the images during training`

bendidi:
as seen in #805 and #724

In some cases, the dataset is not complete or there are some errors in one of its elements, and it's a shame to just throw an error and stop the training for just one buggy element in a dataset.

this change is to log information about the error (to be able to check the dataset for the buggy element) and then continue the training


maskrcnn_benchmark/engine/trainer.py
```diff
@@ -55,6 +55,10 @@ def do_train(
     start_training_time = time.time()
     end = time.time()
     for iteration, (images, targets, _) in enumerate(data_loader, start_iter):
+        
+        if any(len(target) < 1 for target in targets):
+            logger.error(f"Iteration={iteration + 1} || Image Ids used for training {_} || targets Length={[len(target) for target in targets]}" )
```
fmassa: This is python-3 only.

But well, I suppose that with APEX we now have a hard-dependency on Python3 anyway

PR #842, opened 2019-05-31
Save labels in output directory

miguelvr:
Once the training dataset is established, we should save the categories mapping to the model output directory to facilitate deployment.

I'll follow this PR with a general purpose predictor fully initialized from the output directory of the model (containing: model weights, model config (#835) and model labels)


miguelvr:
> LGTM.
> 
> Can you fix the merge conflicts?

should be good to go!

PR #843, opened 2019-05-31
 remove duplication codes in heads' pooler

lsrock1:
Instead of making pooler class directly in each head, use builder function in poolers.py to remove duplicated loading config code.


fmassa:
Hi,

Did you close the PR because it doesn't work for all models?

lsrock1:
> Hi,
> 
> Did you close the PR because it doesn't work for all models?

I made a mistake in coding and i am not sure editing this part is acceptable.
Is this good? Then i will pull request again.

fmassa:
If it works in all the current models, then why not?

But the pooler involves generally not that many lines of code, so maybe keeping it explicit as it currently is might just be simpler? I think I forgot to remove the `make_pooler` when I released the code

lsrock1:
@fmassa make_pooler is only used in backbone/fbnet.py
If consistency is needed, removing make_pooler or using make_pooler is good. What do you think?

PR #853, opened 2019-06-03
the following changes should allow ppl train with negative examples

qianyizhang:
I did similar changes and successfully trained my personal dataset with mask-rcnn.
Have not tested with keypoint branches, but it doesn't raise any errors


qianyizhang:
common request by many ppl #169

PR #863, opened 2019-06-05
Update predictor.py

KelvinCPChiu:
Adding weight loading from saved pth file.


botcs:
Hey @KelvinCPChiu 

Thanks for the PR :) I've been using it for quite long, but now we can merge it as well

zimenglan-sysu-512:
why need it? since the `MODEL.WEIGHT` can do the thing.

Issue #819, opened 2019-05-24
How to test each saved weight?


xiaohai12:
## ❓ Questions and Help
After I saved several weights  from training, I would like to evaluate each saved weights just in case overfitting.
When I tried to run **python -m torch.distributed.launch --nproc_per_node=$NGPUS /path_to_maskrcnn_benchmark/tools/test_net.py --config-file "configs/e2e_mask_rcnn_R_50_FPN_1x.yaml" TEST.IMS_PER_BATCH 16**
it only evaluated the last saved weights.
So, is it possible to evaluate each saved weights ?
By the way,  it costed around 20 mins to evaluate 8000 images on GPU V100,
is it normal or not?

leijiezhang:
Yes the code only evaluate the final check out point, the codes is in /engine/inference.py
if you want to check out other saved weights, you have to re-write the codes, it relates to many python files, be careful. I have tried however failed

xiaohai12:
> Yes the code only evaluate the final check out point, the codes is in /engine/inference.py
> if you want to check out other saved weights, you have to re-write the codes, it relates to many python files, be careful. I have tried however failed

Actually, I checked what is the context inside the last checkpoint file and I found that it was just the path of the saved model weights, which means we only need to modified the last checkpoint file context by using the path of the model weights you want to test.

bendidi:
I haven't tested it, but you can probably just modify the `MODEL.WEIGHT` in the config file you used with the full path to the model you want to test

wangg12:
In `maskrcnn_benchmark/utils/checkpoint.py`, it will always load the last checkpoint in `load()`.
I think you need to change the code to support evaluating on other checkpoints.

wangg12:
Now you can test with any checkpoint you saved.

xiaohai12:
> Now you can test with any checkpoint you saved.

Thanks! It is very useful!

PR #874, opened 2019-06-10
allow test for any checkpoint

wangg12:
Regarding this issue #819


Johnqczhang:
This change isn't backward-compatible when testing a pre-trained model specified in the `.yaml` file as the only argument `--config-file` passing to `test_net.py`, which makes the `checkpointer` fail to load the model file (despite it really exists) and initialize the model from scratch.

Adding a simple line of code can solve this problem, which is given in #894.

PR #881, opened 2019-06-11
Update segmentation_mask.py

Sekunde:
number of instances should be in the first dimension.


davemq:
Sorry, I likely meant #1053. Undo the previous change and try this one.

botcs:
Good catch @Sekunde!

PR #883, opened 2019-06-12
update dockerfile according to the new INSTALL.md

keineahnung2345:
From https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/INSTALL.md:
1. additionally install tqdm, cudatoolkit, apex
2. add FORCE_CUDA argument

Note that the argument CUDA is re-declared, this is done so that its value can be reused, following the instruction [here](https://github.com/moby/moby/issues/37622#issuecomment-412101935).


fmassa:
cc @miguelvr

miguelvr:
LGTM :+1:

PR #890, opened 2019-06-13
add rotation augmentation 

hcx1231:
In some cases, like industrial inspection, rotation is a natural scene as horizontal flip,so I added it to the transform process in training time.Default was set to 0.0.


lsrock1:
Does this work? When i ran this, i got size inconsistent error in cat boxlist.

hcx1231:
> Does this work? When i ran this, i got size inconsistent error in cat boxlist.

I am so sorry.I have resized to keep it the same size with the input box and ran OK  in my own dataset.Let me check it and answer you later.

SkeletonOne:
@hcx1231 Hi, have you tested your implementation on COCO dataset? And do you have an improvement on COCO minival dataset?

ManOfStory:
How can I generate a random rotation angle  by your code? Thanks

PR #903, opened 2019-06-17
Add uncompressed rle condition in Binarymask

lsrock1:
Mask annotation dict can have 2 states
1. compressed RLE that was handled by code.

2. uncompressed RLE, like coco annotation
{'counts': [6, 1, 40, 4, 5, 4, 5, 4, 21], 'size': [9, 10]}
this state was not included by code condition.


botcs:
Hi @lsrock1,
Awesome! thanks for the update.

BTW, I missed uncompressed RLEs so far, did you create them manually?

PR #920, opened 2019-06-22
Update the installation of pycocotools

ChenJoya:
Hi, the pycocotools can be easily installed by pip. Thank you!

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -33,10 +33,7 @@ conda install -c pytorch pytorch-nightly torchvision cudatoolkit=9.0
 export INSTALL_DIR=$PWD
 
 # install pycocotools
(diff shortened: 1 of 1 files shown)


fmassa:
Thanks for the PR!

This seems to be an unofficial pip package. If that's the case, I'd prefer to keep with the current approach for installing pycocotools

botcs:
Indeed, pycocotools has a few PRs merged since the `https://files.pythonhosted.org/packages/96/84/9a07b1095fd8555ba3f3d519517c8743c2554a245f9476e5e39869f948d2/pycocotools-2.0.0.tar.gz` was uploaded, so to keep the latest version it is better to have a fresh clone of the repo

PR #958, opened 2019-07-09
train with background in maskrcnn

abcxs:
I did a comparison test, and I got more accurate results when I tested the background images, although the ap didn't improve.
If there are some bugs, please tell me.


botcs:
Hi @abcxs 

Thanks for the PR.

> I did a comparison test, and I got more accurate results when I tested the background images, although the ap didn't improve.

Could you share your findings?

abcxs:
Surely, i'm doing detection on buildings. For example, buildings don't appear in the mountains, so there are no training background pictures like this during the training, but the test will encounter these pictures, there will be some false check. When I add background pictures to the training, the false check of similar pictures will be greatly reduced.

jeroz1:
I would very much love to see support and more discussion on this. I have a similar issue with negative samples that I like to train to reduce false positives for object detection in aerial images.

PR #964, opened 2019-07-12
Unstructured Pruning Demo

karanchahal:
I was experimenting with pruning of object detection algorithms and was able to get some good results using the detectors in this repo. I though I'd share a popular technique for pruning via a demo here.

The algorithm used for pruning is from the "To Prune or not to Prune" [paper](https://arxiv.org/abs/1710.01878) that is incidentally used by the Tensorflow Model Optimisation Toolkit to perform pruning.

I was able to bring about a sparsity of 70% with almost no loss in mAP using this technique. I suspect getting upto a 90% sparsity is also possible with more training. Here I have only applied pruning masks to the convolutional filters (more specifically around 100 conv layers were pruned).

This branch of research is particularly exciting as one can (theoretically) get a lot of speed ups and compress the network more effectively.  As pytorch is gearing up to support pruning and quantisation, I thought this would be a nice proof of concept. Maybe we can add support for general pruning into this repo by implementing the different algorithms in a subfolder. 

Hope this demo isn't too out of scope for this repo.

Regards,
Karan

### DIFF
--- a/demo/shapes_pruning.ipynb
+++ b/demo/shapes_pruning.ipynb
@@ -1,0 +1,2679 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "shapes-pruning.ipynb",
+      "version": "0.3.2",
+      "provenance": [],
+      "collapsed_sections": [
+        "LUQbRTRocPNN",
+        "aiLvxXRpDbiq",
+        "xnr8tbDz7WjS",
+        "5DC0K7tW7d-M",
+        "BI2ncK7kATEh",
+        "hbzY16ocEdrg",
+        "If8z4OZfDHmC",
+        "mOo-0LGFEAmc",
+        "bbCBInqHFUg7",
+        "tAn3omCjTFGI",
+        "xs_KL1R1aGSA",
+        "NVjPYFN1Pz6D",
+        "tHq9j1HENMMw"
+      ]
+    },
+    "kernelspec": {
+      "name": "python3",
+      "display_name": "Python 3"
+    },
+    "accelerator": "GPU"
+  },
+  "cells": [
+    {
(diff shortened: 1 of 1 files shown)


karanchahal:
Google colab demo [link](https://colab.research.google.com/drive/1Ekik4ZdBYewBRQWSENvRL08KYLz15Hpw)

Darshan2701:
@karanchahal Thanks for an awesome work.
Quick question - Can this be used to prune just the Faster RCNN model? Using resnet50-FPN as the backbone?

karanchahal:
@Darshan2701 yes it definitely can. The good thing about this algorithm is that it's pretty general. Anything with a conv layer is prunable. Let me know it works out okay :)

Darshan2701:
@karanchahal Thanks for the information, Karan. I will try to implement this and will let you know.

92ypli:
@karanchahal Thanks for an awesome work.
Quick question - Can this be used to prune just the MobilenetV2 model?

karanchahal:
Yes it should work. Try it let me know !

enoceanwei:
> Google colab demo [link](https://colab.research.google.com/drive/1Ekik4ZdBYewBRQWSENvRL08KYLz15Hpw)

Dear Karanchahal,

First of all, I am very interesting your jobs, it is an awesome work.

I have followed the codes and try to run it in my local machine.,However, my machine can not resolute the below code:

"
BBoxAugCollator() 
"

In other words, I can not find the "BBoxAugCollator" API in my env libraries, could you tell me how to add the API or solve the problem.

my envs is : win10 x64,  conda python 3.5, cuda10.0, cudnn7.1, pytorch=1.1.0 torchvision=0.2.1

I am looking forward to your reply, many thanks.

Regards

Wei

NikolasBielski:
@karanchahal have you used this for Matterport's MaskRCNN?

PR #983, opened 2019-07-20
Add arg for video capture device choice & update docker docs for webcam demo

slothkong:
I experienced two minor issues when running the [webcam.py](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/demo/webcam.py) demo within docker:

1. I noticed that the index used by `cv2.VideoCapture()` is hard coded to 0. This didn't work for my system. So I had to change the index in the source code (to -1), then copy the new script inside the docker image and spin a new container. To avoid the hassle I expanded the arguments `parser` with a  `--cam-index` option so there is no need to touch the code when the default device mapping doesn't work; instead people can test around different values for the argument .
2. The demo [README.md](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/demo/README.md)  didn't include details on how to run the docker container with GPU support or if it works on Ubuntu 18.04, which is exactly what I tested. So I expanded the documentation with what I tested and worked for me.

### DIFF
--- a/demo/README.md
+++ b/demo/README.md
@@ -42,5 +42,20 @@ docker run --rm -it \
     --config-file configs/caffe2/e2e_mask_rcnn_R_50_FPN_1x_caffe2.yaml
 ```
 
-**DISCLAIMER:** *This was tested for an Ubuntu 16.04 machine, 
+For an image built with nvidia support run:
+
+```
+docker run --rm -it \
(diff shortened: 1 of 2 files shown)


slothkong:
This is a minor improvement to the documentation and the webcam tutorial. Shouldn't be easy to verify. Can we get someone assigned to this PR?

PR #988, opened 2019-07-23
simplified setup routine

dominikandreas:
most people who already have the latest torchvision, torch, apex and pycocotools installed can now just do
`pip install git+https://github.com/facebookresearch/maskrcnn-benchmark`


botcs:
Hey @dominikandreas,
This PRs seems to be an awesome idea.

I'm not so familiar with the version tags, but I think if somebody is only using the lib for demo and so, probably he would aim for a stable version, like [v0.1](https://github.com/facebookresearch/maskrcnn-benchmark/releases/tag/v0.1). 
Do you know how to pip install from github with specific version tags?

Thanks,
Csabi

PR #1022, opened 2019-08-06
change dim=-1 to dim=1 in bounding_box.py for the purpose of tracing using jit

zimenglan-sysu-512:


botcs:
Hi @zimenglan-sysu-512 ,
Thanks for spotting this!

Can you please provide a short script to reproduce this error?

PR #1026, opened 2019-08-08
Add new VoVNet backbone networks into README.md

stigma0617:
we propose an energy and computation efficient backbone network, called VoVNet, which is better and faster than ResNet backbone in terms of accuracy and speed both.
We adopt VoVNet into the maskrcnn-benchmark.


botcs:
Hi @stigma0617,

Could you do please a comparison with the baseline models regarding speed and accuracy performance?

Seeing ResNet-50, -101, and [lightweight baselines](https://github.com/facebookresearch/maskrcnn-benchmark/blob/master/MODEL_ZOO.md#light-weight-model-baselines) would be promising.

Anustup900:
Hey @stigma0617 @botcs , I have worked in this comparisons between Res Net 50 -101 before with Tensorflow , love to work in this issue , can it be assigned to me ?

Issue #725, opened 2019-04-29
IndexError: list index out of range


langjiliang:
## ❓ Questions and Help
Hello, I have a problem. I encountered an error while training the data set. Please ask how to solve this problem. Thank you in advance.
Traceback (most recent call last):
  File "tools/train_net.py", line 186, in <module>
    main()
  File "tools/train_net.py", line 179, in main
    model = train(cfg, args.local_rank, args.distributed)
  File "tools/train_net.py", line 85, in train
    arguments,
  File "/home/sineva/github/maskrcnn-benchmark/maskrcnn_benchmark/engine/trainer.py", line 57, in do_train
    for iteration, (images, targets, _) in enumerate(data_loader, start_iter):
  File "/home/sineva/anaconda3/envs/maskrcnn_benchmark/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 582, in __next__
    return self._process_next_batch(batch)
  File "/home/sineva/anaconda3/envs/maskrcnn_benchmark/lib/python3.7/site-packages/torch/utils/data/dataloader.py", line 608, in _process_next_batch
    raise batch.exc_type(batch.exc_msg)
IndexError: Traceback (most recent call last):
  File "/home/sineva/anaconda3/envs/maskrcnn_benchmark/lib/python3.7/site-packages/torch/utils/data/_utils/worker.py", line 99, in _worker_loop
    samples = collate_fn([dataset[i] for i in batch_indices])
  File "/home/sineva/anaconda3/envs/maskrcnn_benchmark/lib/python3.7/site-packages/torch/utils/data/_utils/worker.py", line 99, in <listcomp>
    samples = collate_fn([dataset[i] for i in batch_indices])
  File "/home/sineva/anaconda3/envs/maskrcnn_benchmark/lib/python3.7/site-packages/torch/utils/data/dataset.py", line 85, in __getitem__
    return self.datasets[dataset_idx][sample_idx]
  File "/home/sineva/github/maskrcnn-benchmark/maskrcnn_benchmark/data/datasets/coco.py", line 83, in __getitem__
    masks = SegmentationMask(masks, img.size, mode='poly')
  File "/home/sineva/github/maskrcnn-benchmark/maskrcnn_benchmark/structures/segmentation_mask.py", line 467, in __init__
    self.instances = PolygonList(instances, size)
[...]

PR #1053, opened 2019-08-20
Fix tensor type checking for pytorch 1.2

liyz15:
Fix https://github.com/facebookresearch/maskrcnn-benchmark/issues/725.
In pytorch 1.2, comparison operations return torch.bool instead of torch.uint8, which causes the error.
Ideally we should change all mask to torch.bool. Indexing with torch.uint8 also leads to infinite warning.

### DIFF
--- a/maskrcnn_benchmark/structures/segmentation_mask.py
+++ b/maskrcnn_benchmark/structures/segmentation_mask.py
@@ -454,7 +454,8 @@ class PolygonList(object):
         else:
             # advanced indexing on a single dimension
             selected_polygons = []
-            if isinstance(item, torch.Tensor) and item.dtype == torch.uint8:
+            if isinstance(item, torch.Tensor) and \
+                    (item.dtype == torch.uint8 or item.dtype == torch.bool):
                 item = item.nonzero()
                 item = item.squeeze(1) if item.numel() > 0 else item
                 item = item.tolist()


zimenglan-sysu-512:
nice！

zimenglan-sysu-512:
if u try to replace `torch.uint8` with` torch.bool`, u should be aware of the evaluation in `coco_eval.py`, in which it needs to set like this:
rles = [
mask_util.encode(np.array(mask[0, :, :, np.newaxis], **dtype=np.uint8**, order="F"))[0]
for mask in masks
]

koenvandesande:
I ran into this one as well, ended up with the same fix. Please merge :)

botcs:
Hi,

Thanks @liyz15 for the fix, I think many of us had this locally, but were too lazy to do a PR :)) so I really appreciate the effort.

@zimenglan-sysu-512 Thank you for pointing this out, I have spent a few hours trying to find out which part I messed up by converting all `uint8` to `bool`.

Contacted @fmassa about PRs waiting for review, hopefully we could move the whole project to use pytorch 1.2 and rewrite the INSTALL.md accordingly soon.

botcs:
Hi,

I've got permission to merge PRs, so we can start updating the lib to use PyTorch v1.2 (or 1.3?)
There are multiple spots in the source where `torch.uint8` is used, and it would be definitely necessary to check if the MODEL_ZOO.md results can be reproduced precisely with higher PyTorch versions.

ethanweber:
Thanks for the fix!

ASONG0506:
> if u try to replace `torch.uint8` with` torch.bool`, u should be aware of the evaluation in `coco_eval.py`, in which it needs to set like this:
> rles = [
> mask_util.encode(np.array(mask[0, :, :, np.newaxis], **dtype=np.uint8**, order="F"))[0]
> for mask in masks
> ]
> for more information, please see [this](https://github.com/matterport/Mask_RCNN/issues/387).

In my case, I did replace torch.uint8 to torch.bool in segmentation_mask.py  and I also modified coco_eval.py, in which I added **dtype=np.uint8**
And it really worked at the begining, but after around 20 iterations, it still occured the same error "list index out of range ", so I am really confused now, can you provide some help?

PR #1065, opened 2019-08-28
Made MaskR-CNN to work with rotated bounding boxes

mrlooi:
I've created a separate repo for MaskRCNN that works with rotated bounding boxes - it performs almost as well as MaskRCNN on COCO but strongly outperforms in datasets where there are significant bounding box overlaps e.g. pencils/pens.

The repo is updated with the latest master branch. 

Additional features: 
- Maskscoring RCNN code (improvements over the original repo. works on both regular and rotated bbox)
- Soft NMS (for both regular and rotated bbox)

I understand if there are too many changes to merge, so I was wondering if instead this repo could be referenced in "Projects using maskrcnn-benchmark" section in README.md? I think it could be a useful repo for people who deal with objects with bounding box overlaps


botcs:
Hey!

This PR seems to be awesome, but before merging it would be nice to see a cleaned up version that has max 5-6 commits below.

Could you open a new PR once its done?

Thank you,
Csabi

PR #1093, opened 2019-09-17
[WIP] Pytorch v1.2

botcs:
Main issue is `uint8` / `bool`.
unit tests are passing, except checkpointer test: had to modify since the expected behavior of `nn.DataParallel` has changed.

Feel free to add any modifications

### DIFF
--- a/maskrcnn_benchmark/modeling/balanced_positive_negative_sampler.py
+++ b/maskrcnn_benchmark/modeling/balanced_positive_negative_sampler.py
@@ -54,10 +54,10 @@ class BalancedPositiveNegativeSampler(object):
 
             # create binary mask from indices
             pos_idx_per_image_mask = torch.zeros_like(
-                matched_idxs_per_image, dtype=torch.uint8
+                matched_idxs_per_image, dtype=torch.bool
             )
             neg_idx_per_image_mask = torch.zeros_like(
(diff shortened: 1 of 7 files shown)


qizhuli:
I was wondering if it would make more sense to dynamically check pytorch version, and choose torch.bool or torch.uint8, so that the code remains backward compatible, and fewer nasty surprises when updating codebase/pytorch? Could be done with a simple helper function `mask_dtype()`:

```python
import torch
from maskrcnn_benchmark.somewhere import get_version

def mask_dtype():
    major, minor = get_version(torch)[:2]
    if major >= 1 and minor >= 2:
        return torch.bool
    else:
        return torch.uint8
```
and put this helper function in `somewhere`:
```python
def get_version(module):
    version = getattr(module, '__version__').split('.')
    parsed_version = []
    for v in version:
        try:
            v = int(v)
            parsed_version.append(v)
        except ValueError:
            parsed_version.append(-1)
    return parsed_version
```
This can also be used to dynamically control the place we call `scheduler.step()`, whose behaviour changed in pytorch 1.1.

botcs:
I like the idea, but I think the same amount of effort would be more useful in adding a few convenience features in detectron2.

There are a lot of stuff still here waiting for cleanup, also detectron2

PR #1095, opened 2019-09-17
Move horizontal flip probability to config

rodrigoberriel:
Simple change to make it more consistent.
It was already there, but `#` :(


botcs:
Nice catch! @rodrigoberriel 

@fmassa any idea why was it commented out in the first case?

PR #1111, opened 2019-09-27
bugfix: use correct config for tta and device handling during inference

bernhardschaefer:
In `compute_on_dataset()` the default cfg object is used instead of the actual experiment configuration.


botcs:
Amazing catch,
thanks!

